PR #1, opened 2023-04-18
Initialize schema with simple build automation

jack-berg:
cc @open-telemetry/configuration-maintainers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,22 @@
+SCHEMA_FILES := $(shell find . -type f -path './schema/*.json' -not -path './.github/*' -not -path './node_modules/*' | sort)
+
+.PHONY: all
+all: install-tools compile-schema validate-example
+
+.PHONY: compile-schema
+compile-schema:
+	@if ! npm ls ajv-cli; then npm install; fi
+	@for f in $(SCHEMA_FILES); do \
+	    npx --no ajv-cli compile --spec=draft2020 -s $$f \
+	        || exit 1; \
+	done
+
+.PHONY: validate-example
+validate-example:
+	@if ! npm ls ajv-cli; then npm install; fi
+	npx --no ajv-cli validate --spec=draft2020 --errors=text -s ./schema/opentelemetry_configuration.json -d ./kitchen-sink-example.yaml
+
+.PHONY: install-tools
+install-tools:
+	npm install
+	@echo "All tools installed"
--- a/kitchen-sink-example.yaml
+++ b/kitchen-sink-example.yaml
@@ -1,0 +1,2 @@
+# The file format version
+file_format: 0.1
--- a/kitchen-sink-example.yaml
+++ b/kitchen-sink-example.yaml
@@ -1,2 +1,2 @@
 # The file format version
-file_format: 0.1
+file_format: "0.1"
--- a/schema/opentelemetry_configuration.json
+++ b/schema/opentelemetry_configuration.json
@@ -1,0 +1,15 @@
+{
+    "$id": "https://opentelemetry.io/otelconfig/OpenTelemetryConfiguration",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "OpenTelemetryConfiguration",
+    "type": "object",
+    "additionalProperties": true,
+    "properties": {
+        "file_format": {
+            "type": "number"
+        }
+    },
+    "required": [
+        "file_format"
+    ]
+}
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -0,0 +1,22 @@
...
+	@if ! npm ls ajv-cli; then npm install; fi
+	@for f in $(SCHEMA_FILES); do \
+	    npx --no ajv-cli compile --spec=draft2020 -s $$f \
+	        || exit 1; \
+	done
+
+.PHONY: validate-example
+validate-example:
+	@if ! npm ls ajv-cli; then npm install; fi
+	npx --no ajv-cli validate --spec=draft2020 --errors=text -s ./schema/opentelemetry_configuration.json -d ./kitchen-sink-example.yaml
```
jack-berg, line 17: It seems useful to not only validate that that the schema files compile, but that one or more canonical example files successfully validate against the schema. The example files can evolve along side the schema.

MikeGoldsmith (approved):
Looks good to me 👍🏻

Makefile
tsloughter: I suppose not necessary at this stage but best to test all of the examples so none can get out of sync with the schema.

schema/opentelemetry_configuration.json
```diff
@@ -0,0 +1,15 @@
+{
+    "$id": "https://opentelemetry.io/otelconfig/OpenTelemetryConfiguration",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
```
tsloughter, line 3: Is this just using the latest or is there something necessary in 2020-12 that is known to be needed eventually?

Curious since existing Erlang json schema lib only supports up to draft 6, so choosing the minimum required version would be nice, otherwise we have to add support for the latest for potentially something that works under an older draft.

Makefile
jack-berg: Yup when there are more examples we should extend the validation.

schema/opentelemetry_configuration.json
jack-berg: Here are the release notes for the versions that have been released after draft 6:

- [Draft 07](https://json-schema.org/draft-07/json-schema-release-notes.html)
- [2019-09](https://json-schema.org/draft/2019-09/release-notes.html)
- [2020-12](https://json-schema.org/draft/2020-12/release-notes.html)

The versions appear to not always be backwards compatible, so an implementation that supports 2020-12 wouldn't necessarily support Draft 06 as well. Somebody fact check me on this if you know better. 

If this turns out to be true, then just as erlang only supports up to draft 6, we'll encounter other implementations that don't support versions less than 2020-12. 

 Not sure which features of JSON Schema we'll want to end up using. However, while I can't pin down exactly when draft 6 was released, its fair to say its quite old and I'm inclined to use the most recent stable version. It looks like there is an effort [stabilize JSON Schema](https://json-schema.org/blog/posts/future-of-json-schema) in the next release, which would help alleviate the problem since there would be a strong incentive for implementations to update to the stable version.

tsloughter: Ah ok, if there isn't backwards compatibility (and I bet that is true since they are drafts) then trying to pin down a "minimum version" doesn't make sense.

Hopefully stabilization happens soon :)

tsloughter (approved):

schema/opentelemetry_configuration.json
```diff
@@ -0,0 +1,15 @@
...
+    "title": "OpenTelemetryConfiguration",
+    "type": "object",
+    "additionalProperties": true,
+    "properties": {
+        "file_format": {
+            "type": "number"
```
tsloughter, line 9: Oops, was doing a last pass over and noticed this. Should this not be a string?

jack-berg: `file_format` refers to the version, e.g. `0.1`.

tsloughter: Right, I mean do we not want a string instead of a float for a version number? Even if only using 2 digits instead of 3 I'd expect a string.

kitchen-sink-example.yaml
```diff
@@ -0,0 +1,2 @@
+# The file format version
+file_format: "0.1"
```
jack-berg, line 2: @tsloughter one thing to note is if this is a string we have to double quote it (or use the `!!str` syntax) else it will be interpreted as a number.

tsloughter: Or use 3 digits, 0.1.0, I think?

jack-berg: This was discussed a bit [here](https://github.com/MrAlias/otel-schema/issues/15). Obviously everything is experimental and still subject to change. Can we merge as is and open an issue to discuss whether a patch version is needed?

PR #3, opened 2023-04-23
Add provider placeholders

jack-berg:
Fill in details of providers in subsequent PRs.

This PR proposes we establish the providers at the top level, rather than nesting under an "sdk" type. I propose we keep everything defined a the top level unless we anticipate enough properties being defined that we'd benefit from additional nesting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SCHEMA_FILES := $(shell find . -type f -path './schema/*.json' -not -path './.github/*' -not -path './node_modules/*' | sort)
+SCHEMA_FILES := $(shell find .  -path './schema/*.json' -exec basename {} \; | sort)
 
 .PHONY: all
 all: install-tools compile-schema validate-example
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -7,14 +7,14 @@ all: install-tools compile-schema validate-example
 compile-schema:
 	@if ! npm ls ajv-cli; then npm install; fi
 	@for f in $(SCHEMA_FILES); do \
-	    npx --no ajv-cli compile --spec=draft2020 -s $$f \
+	    npx --no ajv-cli compile --spec=draft2020 -s ./schema/$$f -r "./schema/!($$f)" \
```
jack-berg, line 10: This change makes all the schemas in the `/schema/` dir (excluding the schema being compiled) available for reference by the schema being compile.

codeboten (approved):

tsloughter (approved):

Makefile
```diff
@@ -1,4 +1,4 @@
-SCHEMA_FILES := $(shell find . -type f -path './schema/*.json' -not -path './.github/*' -not -path './node_modules/*' | sort)
+SCHEMA_FILES := $(shell find .  -path './schema/*.json' -exec basename {} \; | sort)
```
MrAlias, line 1:
```suggestion
SCHEMA_FILES := $(shell find . -path './schema/*.json' -exec basename {} \; | sort)
```

MrAlias (approved):

PR #4, opened 2023-04-25
Attribute limits

jack-berg:
Define general attribute limits, span limits, and log record limits.


kitchen-sink-example.yaml
```diff
@@ -1,11 +1,58 @@
...
+  # Set the max attribute count.
+  #
+  # Environment variable: OTEL_ATTRIBUTE_COUNT_LIMIT
+  attribute_count_limit: 128
+
 # Configure logger provider.
-logger_provider: {}
+logger_provider:
+  # Configure the log record limits. See also attribute_limits.
+  log_record_limits:
```
tsloughter, line 18: Going to be a stickler and question if this should be `logrecord_limits` or not since the environment variables use `LOGRECORD` and the spec talks about `logrecord` :)

tsloughter (approved):

kitchen-sink-example.yaml
```diff
@@ -1,11 +1,58 @@
...
+    # Set the max log record attribute value size. Overrides attribute_limits.attribute_value_length_limit.
+    #
+    # Environment variable: OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT
+    attribute_value_length_limit: 4096
```
codeboten, line 22: since these are already under `_limits`, does it make sense to remove the `_limit` suffix?

tsloughter: I don't know why but also think I prefer having the `_limit` suffix.

codeboten: i'm ok with keeping the suffix for now, it can always be revisited later

kitchen-sink-example.yaml
```diff
@@ -1,11 +1,58 @@
 # The file format version
 file_format: "0.1"
 
+# Configure general attribute limits. See also tracer_provider.span_limits, logger_provider.log_record_limits.
```
codeboten, line 4:
```suggestion
# Configure general attribute limits. See also tracer_provider.span_limits, logger_provider.logrecord_limits.
```

kitchen-sink-example.yaml
codeboten, line 18:
```suggestion
  logrecord_limits:
```

schema/logger_provider.json
```diff
@@ -5,5 +5,18 @@
     "type": "object",
     "additionalProperties": false,
     "properties": {
+        "log_record_limits": {
```
codeboten, line 8:
```suggestion
        "logrecord_limits": {
```

codeboten (approved):

PR #14, opened 2023-05-05
Add propagators configuration

jack-berg:
I considered making each propagator an object to allow future propagators to include configuration options:
```
propagators:
  foo_propagator:
    key: value
```
But chose against it because none of the current propagators accept configuration, and we can evolve into this future if needed with something like:
```
# Entries provide configuration options for .propagators entires.
propagator_config:
  foo_propagator:
    key: value
propagators: [foo_propagator]
```

### DIFF
--- a/kitchen-sink-example.yaml
+++ b/kitchen-sink-example.yaml
@@ -28,6 +28,11 @@ logger_provider:
 # Configure meter provider.
 meter_provider: {}
 
+# Configure context propagators.
+#
+# Environment variable: OTEL_PROPAGATORS
+propagators: [tracecontext, baggage, b3, b3multi, jaeger, xray, ottrace]
+
 # Configure tracer provider.
 tracer_provider:
   # Configure span limits. See also attribute_limits.
--- a/schema/opentelemetry_configuration.json
+++ b/schema/opentelemetry_configuration.json
@@ -27,6 +27,12 @@
         "meter_provider": {
             "$ref": "meter_provider.json"
         },
+        "propagators": {
+            "type": "array",
+            "items": {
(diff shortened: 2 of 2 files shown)


tsloughter:
I'm pretty sure I brought this up at the time `OTEL_PROPAGATORS` was defined, so there may be an answer to this question already and I'll try to find it, but these are all text map propagators, what happens when binary propagators are added?

The SDK having to check each propagator's type and create multiple composite propagators if there are propagators of multiple types would work (at least in languages I can think of) -- and thus requiring that if `tracecontext` adds a binary format it is named `tracecontext_bin` or simiar -- but feels a little off, so wanted to bring it up.

codeboten (commented):
@tsloughter i suppose we could add an optional `format` field to the propagator configuration to support this, and it defaults to `text`? although i guess we could also add this later

tsloughter:
Yea, that'd work, esp good since it can be added later.

Just feels weird to define a list `propagators` that is actually defining the global lists of propagators for N types instead of it being `text_map_propagators`.

jack-berg:
Yeah the `propagators` field is definitely currently defining a [composite text map propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#composite-propagator). When binary propagators are added we could add a new `binary_propagators` key? Would need to do some validation on the array entries to ensure that they are text map or binary, respectively, based on whether they're referenced in `propagators` or `binary_propagators`.

MrAlias (approved):

codeboten (approved):
I agree that we can solve this in the future.

PR #15, opened 2023-05-05
adding meter provider configuration

### LINKED ISSUE #9: Configure meter provider

codeboten:
Using anchors, folks will be able to re-use configuration that's using standard YAML functionality. This differs from the configuration in the OTEP where exporters were defined within each provider and used via their names.

Note the section where exporters anchor is defined at the top level are not validated, the json schema validation is done where the anchor is used.

Fixes #9


kitchen-sink-example.yaml
```diff
@@ -26,7 +61,45 @@ logger_provider:
     attribute_count_limit: 128
 
 # Configure meter provider.
-meter_provider: {}
+meter_provider:
+  metric_readers:
+    - type: prometheus
```
tsloughter, line 66: Should this be `pull`?

Was the issue of readers/exporters being so tightly bound in their definitions discussed in the otel-schema repo or one of the meetings? I feel like in the spec there was discussion about potentially having users just configure exporters and the Reader be sort of figured out from that...

Defining the exporter and the Reader being "figured out" from that is I guess sort of what is being done if `type` can be `prometheus`, instead of type being `pull` and `prometheus` being an exporter defined under that. But then shouldn't the same be done for `otlp`?

kitchen-sink-example.yaml
```diff
@@ -1,5 +1,40 @@
 # The file format version
 file_format: "0.1"
+exporters: &otlp-exporter
+  otlp:
+      # Sets the protocol.
+      #
+      # Environment variable: OTEL_EXPORTER_OTLP_PROTOCOL, OTEL_EXPORTER_OTLP_METRICS_PROTOCOL
+      protocol: http/protobuf
```
jack-berg, line 8: I think the keys here are indented an extra two spaces?

kitchen-sink-example.yaml
jack-berg: By saying that the `type` is prometheus, we're more firmly asserting that for the purposes of file configuration, prometheus is a specific pull based reader, not an exporter.

> I feel like in the spec there was discussion about potentially having users just configure exporters and the Reader be sort of figured out from that...

Not defining the reader means we need to have somewhere to define the reader properties, like the interval for periodic metric reader. Feels wrong to put this on the exporter configuration, since it would result in metric exporter configuration being quite different from trace and log exporter configuration.

tsloughter: By using a type of `prometheus` it is already sort of merging the two (reader and exporter) but instead of defining the exporter the user defines the reader. 

Maybe that makes the most sense and there would be `type: otlp`?

codeboten: Hmmm.... i agree that the asymmetry here is somewhat confusing. Thinking in the context of what users have the ability to configure today w/ the `OTEL_METRICS_EXPORTER` [environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#exporter-selection) which support `otlp` and `prometheus`:

- `otlp` configures a periodic push based metric reader with OTLP as the exporter
- `prometheus` configures a prometheus metric reader

If we're trying to make the underlying results clear to end users, then the current configuration is "correct", although not really clear. Adding `type: otlp` would be ok if we supported a pull-based OTLP metric reader, but as of now that's not the case. Similarly, if anyone was to add a prometheus remote write exporter, when i would expect the exporter configuration to live under a push-based metric reader.

tsloughter: > Adding `type: otlp` would be ok if we supported a pull-based OTLP metric reader, but as of now that's not the case.

I don't understand. Wouldn't the opposite be the case? `otlp` and `prometheus` work as types because they are only pull or push. And new types could of course be added, like `prometheus_remote_write`.

Lining up with the environment variables would also mean configuring the exporter type and not the reader. So maybe like:

```
meter_provider:
  exporters:
    - type: otlp
      interval: 60000
      timeout: 30000
```

```
meter_provider:
  exporters:
    - type: prometheus
      host: "localhost"
      port: 9464
```

Or:

```
meter_provider:
  exporters:
    prometheus:
      host: "localhost"
      port: 9464
```

These seem more inline with what the environment variables are doing. Not that I think this has to stick strictly to the environment variables, but each exporter only works with a specific type of reader, so requiring the user to set that up seems unnecessary and to just add complexity.

jack-berg:
```
meter_provider:     
  readers:
    - type: prometheus
      host: localhost
      port: 9464
    - type: openmetrics
      host: localhost
      port: 9464
    - type: periodic
       exporter: &otlp
       interval: 60000
```

```
meter_provider:
  readers:
    - type: pull
      exporter:
        type: prometheus
        host: localhost
        port: 9464
    - type: pull
      exporter:
        type: openmetrics
        host: localhost
        port: 9465
    - type: periodic
      exporter: &otlp
      interval: 60000
```

codeboten: I updated the PR to follow the second option shown in @jack-berg's comment above. I think calling out the type of reader explicitly in both push and pull makes it clear to the end user what is happening with less magic.

kitchen-sink-example.yaml
codeboten: Fixed, thanks for catching

kitchen-sink-example.yaml
```diff
@@ -26,7 +71,44 @@ logger_provider:
...
 
 # Configure meter provider.
-meter_provider: {}
+meter_provider:
+  metric_readers:
+    # Add a pull-based metric reader.
+    - type: pull
+      args:
+        exporter:
+          <<: *prometheus-exporter
```
jack-berg, line 80: #nit: since this is the only use of the prometheus exporter config, how about we include that directly here?

schema/meter_provider.json
```diff
@@ -5,5 +5,127 @@
...
+                "timeout": {
+                    "type": "integer"
+                },
+                "exporter": {
+                    "$ref": "#/$defs/MetricExporter"
+                },
+                "host": {
+                    "type": "string"
+                },
+                "port": {
```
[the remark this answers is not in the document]
codeboten: Removed 👍🏻

kitchen-sink-example.yaml
codeboten: Expanded the config and removed the anchor

kitchen-sink-example.yaml
```diff
@@ -26,7 +71,44 @@ logger_provider:
...
+    # Add a periodic metric reader.
+    #
+    # Environment variable: OTEL_METRICS_EXPORT_*, OTEL_METRICS_EXPORTER
+    - type: periodic
+      args:
```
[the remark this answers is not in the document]
codeboten: > We could have metric_readers be an object instead of array, with keys which refer to the type, and values whose schema is enforced based on the key

How would this work in the case where a user may want to configure multiple periodic exporters?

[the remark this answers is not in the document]
jack-berg: Same way we were planning on supporting multiple exporters where the keys identify the exporter:
```
metric_readers:
  periodic/foo: ...
  periodic/bar: ...
```
I don't love it... Better if we can find a way to do:
> Maybe the best we can do is find a way to conditionally enforce a schema on args based on the value of type
But if you agree, that can be in a subsequent PR.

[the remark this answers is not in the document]
codeboten: Oh right.... yes that's fine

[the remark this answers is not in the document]
codeboten: I've updated the schema to follow @jack-berg's suggestion

tsloughter (approved):
Looks good. Only comment is that it looks like args is an unnecessary key under the reader since its all args to the reader anyway.

PR #16, opened 2023-05-05
Sampler schema

jeremydvoss:
Sampler id is basically the stand in for OTEL_TRACES_SAMPLER's use of entry points. It includes entry points under the opentelemetry_traces_sampler entry point group.

Extra properties are true so that those samplers or sampler factories can take additional arguments.


tsloughter:
I don't think we should restrict the types of things to strings to match environment variables. Unless the schema is being limited to just taking an arg, like `OTEL_TRACE_SAMPLER_ARG`, and not defining a schema for each builtin sampler type.

schema/sampler.json
```diff
@@ -0,0 +1,15 @@
...
+    "additionalProperties": true,
+    "properties": {
+        "sampler_id": {
+            "type": "string"
+        },
+        "argument": {
+            "type": "string"
+        }
+    }
+}
```
jack-berg, line 15: You'll need to extend [kitchen-sink-example.yaml](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/kitchen-sink-example.yaml) to show example usage.

Also, sampler will be defined as a part of [trace_provider](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.json), so you'll need to extend that schema to include the appropriate fields.

jack-berg (commented):
We've already done some thinking about sampler configuration. Check out [this example configuration](https://github.com/MrAlias/otel-schema/blob/main/config.yaml#L178-L229) for the types of things that need to be configurable and our first attempt at structuring that info.

schema/sampler.json
jeremydvoss: I believe I have addressed this now.

jeremydvoss:
> I don't think we should restrict the types of things to strings to match environment variables. Unless the schema is being limited to just taking an arg, like `OTEL_TRACE_SAMPLER_ARG`, and not defining a schema for each builtin sampler type.

With better understanding from the last sig, I no longer need sampler arg. the samplers arguments will be represented by the additional properties.

kitchen-sink-example.yaml
```diff
@@ -112,6 +112,26 @@ propagators: [tracecontext, baggage, b3, b3multi, jaeger, xray, ottrace]
 
 # Configure tracer provider.
 tracer_provider:
+  sampler:
+    # Choose sampler id. Must be entry point name of sampler
+    sampler_id: parent_based
```
jeremydvoss, line 117: I decided sampler_id is clearer than "type" which we used in our weekly sig @jack-berg Thoughts?

jack-berg:
Superseded by #24.

PR #18, opened 2023-06-23
add span processor + exporter to tracer provider

codeboten:
This PR adds configuration for two processors (simple + batch) and span exporters for OTLP and console output.

### DIFF
--- a/schema/common.json
+++ b/schema/common.json
@@ -8,6 +8,52 @@
             "type": "object",
             "additionalProperties": true,
             "title": "Headers"
+        },
+        "Otlp": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "protocol": {
+                    "type": "string",
+                    "pattern": "^(http|grpc)\\/(protobuf|json)"
+                },
+                "endpoint": {
+                    "type": "string"
+                },
+                "certificate": {
+                    "type": "string"
+                },
+                "client_key": {
+                    "type": "string"
+                },
+                "client_certificate": {
+                    "type": "string"
+                },
+                "headers": {
+                    "$ref": "#/$defs/Headers"
+                },
+                "compression": {
+                    "type": "string"
+                },
+                "timeout": {
+                    "type": "integer"
+                },
+                "temporality_preference": {
+                    "type": "string"
+                },
+                "default_histogram_aggregation": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "endpoint",
+                "protocol"
+            ],
+            "title": "Otlp"
+        },
+        "Console": {
+            "type": "object",
+            "additionalProperties": false
         }
     }
 }
--- a/schema/common.json
+++ b/schema/common.json
@@ -37,12 +37,6 @@
                 },
                 "timeout": {
                     "type": "integer"
-                },
-                "temporality_preference": {
-                    "type": "string"
-                },
-                "default_histogram_aggregation": {
-                    "type": "string"
                 }
             },
             "required": [
--- a/schema/tracer_provider.json
+++ b/schema/tracer_provider.json
@@ -59,6 +59,18 @@
             "additionalProperties": false,
             "title": "BatchSpanProcessor",
             "properties": {
+                "schedule_delay": {
+                    "type": "integer"
+                },
+                "export_timeout": {
+                    "type": "integer"
+                },
+                "max_queue_size": {
+                    "type": "integer"
+                },
+                "max_export_batch_size": {
+                    "type": "integer"
+                },
                 "exporter": {
                     "$ref": "#/$defs/SpanExporter"
                 }
@@ -67,12 +79,19 @@
         "SpanExporter": {
             "type": "object",
             "additionalProperties": true,
-            "patternProperties": {
-                "^otlp.*": {
+            "minProperties": 1,
+            "maxProperties": 1,
+            "properties": {
+                "otlp": {
                     "$ref": "common.json#/$defs/Otlp"
                 },
-                "^console.*": {
+                "console": {
                     "$ref": "common.json#/$defs/Console"
+                }
+            },
+            "patternProperties": {
+                ".*": {
+                    "type": "object"
                 }
             }
         }
--- a/schema/tracer_provider.json
+++ b/schema/tracer_provider.json
@@ -5,6 +5,18 @@
     "type": "object",
     "additionalProperties": false,
     "properties": {
+        "processors": {
+            "type": "object",
+            "additionalProperties": true,
+            "patternProperties": {
+                "^batch.*": {
+                    "$ref": "#/$defs/BatchSpanProcessor"
+                },
+                "^simple.*": {
+                    "$ref": "#/$defs/SimpleSpanProcessor"
+                }
+            }
+        },
         "limits": {
             "title": "SpanLimits",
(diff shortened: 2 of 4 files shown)


schema/tracer_provider.json
```diff
@@ -30,5 +42,80 @@
...
+                    "$ref": "#/$defs/Console"
+                }
+            }
+        },
+        "Console": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {}
+        },
+        "Otlp": {
```
aabmass, line 84: Most of this looks in common with the metric OTLP exporter. Not super familiar with JSON schema but it could be nice to reuse some of it so they don't go out of sync

schema/tracer_provider.json
```diff
@@ -30,5 +42,80 @@
                 }
             }
         }
+    },
+    "$defs": {
+        "SimpleSpanProcessor": {
```
aabmass, line 47: What do you think of omitting SimpleSpanProcessor? It's kind of a footgun that a lot of users get hung up on

codeboten: It's definitely an option, i included because its a nice to have when starting, but its not necessary

schema/tracer_provider.json
codeboten: There are only 2 fields which are different, i suppose since those fields are not required, they could be the same definition... only downside i could think of is someone getting prompted to fill in unnecessary fields for a span/log exporter

codeboten: I've combined the otlp + console exporters under common.json, PTAL

schema/tracer_provider.json
tsloughter: Situations like a Lambda function.

Is it a footgun because of the name? Maybe should have been BlockingSpanProcessor or something :)

schema/common.json
```diff
@@ -8,6 +8,52 @@
...
+                    "type": "string"
+                },
+                "timeout": {
+                    "type": "integer"
+                },
+                "temporality_preference": {
+                    "type": "string"
+                },
+                "default_histogram_aggregation": {
+                    "type": "string"
```
jack-berg, line 45: Instead of having these metric specific fields in the common definition, we might consider adding a metric specific Otlp exporter type, and use all of syntax to extend it with the metrics specific fields:
```
{
  "allOf": [{ "$ref": "common.json#/#defs/Otlp" }],
  "properties": {
      "temporality_preference": { "type": "string" },
      "default_histogram_aggregation": { "type": "string" },
  }
}
```
I'm not sure of how well supported `allOf` is among the JSON schema validation libraries, but its worth a shot. If its not well supported, we can document that we try to avoid `allOf` in the README.

schema/tracer_provider.json
```diff
@@ -30,5 +42,39 @@
...
+            }
+        },
+        "BatchSpanProcessor": {
+            "type": "object",
+            "additionalProperties": false,
+            "title": "BatchSpanProcessor",
+            "properties": {
+                "exporter": {
+                    "$ref": "#/$defs/SpanExporter"
+                }
```
jack-berg, line 64: Is the plan to add the [other batch span processor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#batching-processor) parameters in a different PR?

schema/tracer_provider.json
jack-berg: Its a footgun because it triggers one export request per span. Extremely noisy!

aabmass: > Situations like a Lambda function.

They could also force flush their BatchSpanProcessor at the end of the function. At least that's what we've been recommending for GCP serverless. @tsloughter would this not work for Lambda?

> Is it a footgun because of the name? Maybe should have been BlockingSpanProcessor or something :)

The performance issues from blocking the calling thread have been a really problematic trap that new users fall in when they copy code snippets online. Also what Jack said. We should at least be careful to not feature it prominently in tons of examples.

For something like a logging exporter, it would still be nice to at least take serialization and IO off the request thread instead of blocking?

schema/common.json
aabmass: I think I had issues with this when I tried the golang json schema codegen. Worth a try though.

schema/tracer_provider.json
codeboten: The simple span processor is the only other processor currently defined in the specification, for the sake of completeness it should be supported, but I agree that maybe it's just not featured in the "recommended" example. WDYT?

schema/tracer_provider.json
codeboten: Added the remaining params.

schema/common.json
codeboten: `allOf` is not currently supported by the go-jsonschema (library im currently using) https://github.com/omissis/go-jsonschema/issues/6

Doesn't mean we shouldn't use it, but i won't have time to look at it just yet. I could use two different types, or keep the one type w/ all needed fields (since they're optional anyways). Any preference one way or the other?

jack-berg: > allOf is not currently supported by the go-jsonschema (library im currently using) https://github.com/omissis/go-jsonschema/issues/6

Dang.. I'm using all of in sampler config in #21 as well. Any ideas how we could structure the schema to constrain the parameters of the various samplers without using allOf?

codeboten: After looking further into allOf, I'm not sure how well it would work for something like the OTLP exporter. `allOf` requires validation against everything defined in the schema, which means that something like `additionalProperties` must be set to `true` in the `Otlp` definition to allow additional fields for `OtlpMetric`.

The following fails validation because OtlpMetric includes additional fields:

```json
      "Otlp": {
            "type": "object",
         ...
         },
        "OtlpMetric": {
            "type": "object",
            "additionalProperties": true,
            "allOf": [{ "$ref": "common.json#/$defs/Otlp" }],
            "properties": {
                "temporality_preference": { "type": "string" },
                "default_histogram_aggregation": { "type": "string" }
            }
        }
```

Making `additionalProperties` true on `Otlp` removes some beneficial validation, for example if someone makes a typo when entering an expected field.

Maybe the best thing is to keep these two definition separate for now?

jack-berg: > Maybe the best thing is to keep these two definition separate for now?

That's fair. So that means we need to separately address whether we want to allow usage of allOf in other scenarios, like #21.

schema/tracer_provider.json
```diff
@@ -30,5 +42,56 @@
...
+        "SpanExporter": {
+            "type": "object",
+            "additionalProperties": true,
+            "patternProperties": {
+                "^otlp.*": {
+                    "$ref": "#/$defs/OtlpTrace"
+                },
+                "^console.*": {
+                    "$ref": "common.json#/$defs/Console"
+                }
```
jack-berg, line 88: Let's apply the pattern in #23 to this.

codeboten: Done

schema/common.json
codeboten: Managed to get this working through the use of `unevaluatedProperties`, however decided to go back to splitting the otlp metric exporter from the common otlp exporter due to the lack of support for `anyOf` in the tooling. This can always be revisited in the future.

jack-berg (approved):

PR #21, opened 2023-06-26
Add sampler configuration

jack-berg:
Alternative to #16 which imposes type safety on the sampler configuration parameters by using conditional schemas based on the value of the `id` field.

### DIFF
--- a/schema/tracer_provider.json
+++ b/schema/tracer_provider.json
@@ -29,6 +29,89 @@
                     "type": "integer"
                 }
             }
+        },
+        "sampler": {
+            "$ref": "#/$defs/Sampler"
+        }
+    },
+    "$defs": {
+        "Sampler": {
+            "type": "object",
+            "additionalProperties": true,
+            "properties": {
+                "id": {
+                    "type": "string"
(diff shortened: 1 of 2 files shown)


schema/tracer_provider.json
```diff
@@ -29,6 +29,89 @@
...
+                "id": {
+                    "type": "string"
+                }
+            },
+            "allOf": [
+                {
+                    "if": {
+                        "properties": {
+                            "id": {
+                                "const":  "jaeger_remote"
```
jack-berg, line 51: Interpret as: 
- when id == "jaeger_remote", then enforce the relevant schema on endpoint, interval, initial_sampler properties
- when id == "parent_based", then enforce the relevant schema on root, remote_parent_sampled, remote_parent_not_sampled, local_parent_sampled, local_parent_not_sampled properties
- when id == "trace_id_ratio_based", then enforce relevant schema on ratio property

"always_on", and "always_off" don't have entries because they're not configurable.

jack-berg:
Closing in favor of #24.

PR #22, opened 2023-06-26
remove span_, metric_, logrecord_ prefixes

### LINKED ISSUE #20: metric_readers section name could be shortened to readers
### TL;DR

The `metric_readers` section can be shortened to `readers` which would be shorter while still being readable and concrete enough, and would make the OpenTelemetry Collector telemetry configuration cleaner.

### Longer explanation & context

On open-telemetry/opentelemetry-collector/pull/7871 we are using the Configuration WG's `meter_provider` configuration schema as part of the OpenTelemetry Collector telemetry configuration. In this case, configuring a reader looks like the following:

```yaml
# collector.yaml sample telemetry configuration
service:
  telemetry:
    metrics:
      metric_readers:
        pull/prometheus:
          # ...
```

As noted, this matches the example on this repository:

https://github.com/open-telemetry/opentelemetry-configuration/blob/362b23ec9bb61537e6479a9546d7e8f8a388ec81/kitchen-sink-example.yaml#L64-L69

Although I believe in our case it is specially more salient because our top level configuration is named `metrics`, I still think on the base example on this repository the `metric_` prefix for the `metric_readers` subsection is redundant: a meter provider will presumably never have any other kind of reader.

codeboten:
Closes #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,6 @@ Allowable changes:
 The following defines guidelines used to produce configuration schema:
 
 1. To remove redundant information from the configuration file, prefixes for data produced by each of the providers
-will removed from configuration options. For example, under the `meter_provider` configuration, metric readers will be
+will be removed from configuration options. For example, under the `meter_provider` configuration, metric readers will be
 identified by the word `readers` rather than by `metric_readers`. Similarly, the prefix `span_` will be dropped for tracer
 provider configuration, and `logrecord` for logger provider.
--- a/README.md
+++ b/README.md
@@ -30,3 +30,12 @@ Allowable changes:
 
 - For major versions: All changes are permitted.
 - For minor versions: TBD
+
+## Naming guidelines
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -30,3 +30,12 @@ Allowable changes:
 
 - For major versions: All changes are permitted.
 - For minor versions: TBD
+
+## Naming guidelines
+
+The following defines guidelines used to produce configuration schema:
+
+1. To remove redundant information from the configuration file, prefixes for data produced by each of the providers
+will removed from configuration options. For example, under the `meter_provider` configuration, metric readers will be
```
MrAlias, line 39:
```suggestion
will be removed from configuration options. For example, under the `meter_provider` configuration, metric readers will be
```

MrAlias (approved):

mx-psi (approved):

PR #23, opened 2023-06-28
Prevent multiple metric exporters

jack-berg:
Right now there's nothing preventing multiple metric exporters from being registered, i.e.:
```
meter_provider:
  readers:
    periodic/otlp:
      exporter:
        otlp:
          endpoint: ...
        console:
```
This PR adds `minProperties` and `maxProperties` to prevent multiple. Also enforces that unrecognized exporter keys (i.e. custom exporters) must be type `object`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,14 +7,14 @@ all: install-tools compile-schema validate-example
 compile-schema:
 	@if ! npm ls ajv-cli; then npm install; fi
 	@for f in $(SCHEMA_FILES); do \
-	    npx --no ajv-cli compile --spec=draft2020 -s ./schema/$$f -r "./schema/!($$f)" \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,14 +7,14 @@ all: install-tools compile-schema validate-example
 compile-schema:
 	@if ! npm ls ajv-cli; then npm install; fi
 	@for f in $(SCHEMA_FILES); do \
-	    npx --no ajv-cli compile --spec=draft2020 -s ./schema/$$f -r "./schema/!($$f)" \
+	    npx --no ajv-cli compile --spec=draft2020 --allow-matching-properties -s ./schema/$$f -r "./schema/!($$f)" \
```
jack-berg, line 10: See https://ajv.js.org/strict-mode.html#overlap-between-properties-and-patternproperties-keywords

codeboten (approved):

PR #25, opened 2023-06-28
Add view configuration schema

jack-berg:


schema/meter_provider.json
```diff
@@ -124,6 +130,102 @@
...
+                        },
+                        "meter_version": {
+                            "type": "string"
+                        },
+                        "meter_schema_url": {
+                            "type": "string"
+                        }
+                    }
+                },
+                "view": {
```
MrAlias, line 159: This defines the stream configuration, right?

```suggestion
                "stream": {
```

jack-berg: Yes, the vocabulary you add in https://github.com/open-telemetry/opentelemetry-specification/pull/3524 helps clarify this. Will update to stream to reflect that 👍

MrAlias (approved):

PR #27, opened 2023-07-10
Adjust processors and readers to be arrays instead of objects

jack-berg:


MrAlias (approved):

schema/meter_provider.json
```diff
@@ -6,14 +6,20 @@
...
+                "additionalProperties": false,
+                "minProperties": 1,
+                "maxProperties": 1,
+                "properties": {
+                    "periodic": {
+                        "$ref": "#/$defs/PeriodicMetricReader"
+                    },
+                    "pull": {
+                        "$ref": "#/$defs/PullMetricReader"
+                    }
```
codeboten, line 22: is a patternProperties missing for the readers definition?

jack-berg: I don't think so. Every reader should either be a pull reader with an exporter specified as an argument, or a periodic reader (aka push) with an exporter defined as an argument. The way we've structured it, I don't believe there are any additional "types" or readers.

schema/tracer_provider.json
```diff
@@ -6,14 +6,25 @@
...
+                    "batch": {
+                        "$ref": "#/$defs/BatchSpanProcessor"
+                    },
+                    "simple": {
+                        "$ref": "#/$defs/SimpleSpanProcessor"
+                    }
                 },
-                "^simple.*": {
-                    "$ref": "#/$defs/SimpleSpanProcessor"
+                "patternProperties": {
```
codeboten, line 24: Unclear if this is a side-effect of the go-jsonschema library, but this change has the side effect of restricting the types of processors to only simple or batch, where as they were a map of interface objects previously, they are now elements w/ the following definition:

```go

type TracerProviderJsonProcessorsElem struct {
	// Batch corresponds to the JSON schema field "batch".
	Batch *BatchSpanProcessor `mapstructure:"batch,omitempty"`

	// Simple corresponds to the JSON schema field "simple".
	Simple *SimpleSpanProcessor `mapstructure:"simple,omitempty"`
}
```

codeboten: You can see the full change here: https://github.com/open-telemetry/opentelemetry-collector/pull/8046

jack-berg: That's not good. I wonder if there's a way to configure the go-jsonschema library to access additional properties when additionalProperties is set to true.

codeboten: After further investigation, this appears to be a limitation of the library at this point

jack-berg: That's frustrating. Some thoughts:

- This should only impact configuration of custom components (processors, exporters, samplers, etc) since the built in ones will be able to have properties defined in the schema. Will need to configure custom components eventually, but maybe not critical for MVP?
- Can / should open an issue with go-jsonschema if one isn't already open. 
- If the code gen library proves to rough around the edges to be valuable, there's always the option of hand rolling the model structs, but still using off-the-shelf code to parse the YAML to those structs. In java, I suspect we'll end up in this situation eventually since we'll want to tightly control the shape of model classes, but I want to lean on code gen to speed up prototyping.

What do you think?

codeboten: I agree, I'll take an action item to open an issue, and i agree it's likely that code generation may only prove useful in the prototyping phase here.

codeboten (approved):

PR #37, opened 2023-08-17
add enum for instrument types

### LINKED ISSUE #36: instrument type should be restricted to known instruments
Currently the instrument_type field is a string. It would be nice to restrict it to a known list of instruments for validation.

codeboten:
This will allow us to validate instrument types

Fixes #36

### DIFF
--- a/schema/meter_provider.json
+++ b/schema/meter_provider.json
@@ -157,7 +157,16 @@
                             "type": "string"
                         },
                         "instrument_type": {
-                            "type": "string"
+                            "type": "string",
+                            "enum": [
+                                "counter",
+                                "gauge",
+                                "histogram",
+                                "observable_counter",
+                                "observable_gauge",
+                                "observable_updown_counter",
+                                "updown_counter"
+                            ]
                         },
                         "meter_name": {
                             "type": "string"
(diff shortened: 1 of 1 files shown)


tsloughter:
Won't this tie us to not supporting custom instruments in the configuration?

codeboten:
Are custom instruments something we should support? I'm not sure how custom instruments would work w/ the API that lists the supported instruments

jack-berg:
> Are custom instruments something we should support? I'm not sure how custom instruments would work w/ the API that lists the supported instruments

I don't know of any plans / discussions around custom instruments in the otel metrics API, so an enum seems appropriate.

tsloughter:
Yea, nevermind, I was mixing up with custom aggregators.

:+1:

jack-berg (approved):
Slight preference for "*_up_down_*" over "*_updown_*"

codeboten:
Looking at the spec, it appears either is ok. I'll open a PR against the spec to ensure consistency there as well.

codeboten:
@jack-berg please have a look at https://github.com/open-telemetry/opentelemetry-specification/pull/3665

PR #54, opened 2023-10-30
Add Python prototype

ocelotl:
This is a configuration prototype in Python.

AFAIK, other prototypes currently parse the configuration file into a certain, language-specific model, then use `if/else`s to query this model, depending on what is found in the model with this querying actual objects are created.

This approach has a few disadvantages:
1. It requires manual `if/else` querying.
2. It doesn't guarantee a complete implementation of all the possible scenarios that any configuration file may have.

The previous point 2 requires a bit more explanation. For example, a configuration implementation that manually does `if/else` querying of the model may miss a possible scenario that is included in the configuration file but is not implemented. For example, a tracer provider may have `batch` and `simple` span processors, but maybe the implementation only queries the configuration model for `batch`, forgetting about `simple`. For someone implementing the configuration model in this way, it is easy to make this kind of mistake because it is not obvious from any configuration file or the schema which are all the possible configuration scenarios.

This prototype implements configuration in a different way.

It first [resolves the schema](https://github.com/open-telemetry/opentelemetry-configuration/pull/54/files#diff-98128e0e289fa29b3b41f0fdd75a0340ddda4556ff7593f76d1dd66b4ab28d2f), by replacing all the external `$defs` in the schema with their actual definition. This produces a complete _tree_ representation of the schema.

Then it recursively traverses the schema tree to find all the possible paths. With this information, it writes a Python [file](https://github.com/open-telemetry/opentelemetry-configuration/pull/54/files#diff-7585b42b4677df738a440e9c42f6968237ae9dabc70ba0f00867a506c89aa14b) with one function per schema path. The arguments that these functions receive are taken from the schema. All these functions are associated with their schema path in a Python dictionary.

Now, I implemented the content of these functions (so far I have only implemented the content of the functions that are related to `Resource` and `TracerProvider` creation). It is important to mention here that this means that this prototype can also have an incomplete implementation (as mentioned in point 2 above), but in this case it is clear for whoever is implementing this what is missing from the implementation: the functions whose content has not been implemented yet.

Finally, this prototype takes the configuration file, parses it into a Python dictionary (which is another _tree_ representation). It recursively traverses this tree and when it finds a node it this tree it calls the function associated with this node path. For example, to create a resource:

It first finds the `resource` node in the [configuration tree](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L368). It traverses through all the attributes and finds that first it needs to create the object for `attributes`, so it recursively keeps traversing the configuration tree until it finds the `resource/attributes` [node](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L372). Since all the attributes for `resource/attributes` are _terminal_ (strings, ints, floats, booleans), they can be used right away to create the `attributes` object. This object is recursively returned to the `resource` node which can now complete the creation of the `Resource` object which is finally returned.

To try this prototype, install [`nox`](https://nox.thea.codes/en/stable/), go to the `opentelemetry-configuration/prototypes/python` directory and run:

`nox -- -s -vvvv -k test_create --pdb`

This will run [this](https://github.com/open-telemetry/opentelemetry-configuration/pull/54/files#diff-836e622d33f512afe656380a75d1eb34ee990aed487c75927e115fed6cd4c63aR136) test case which uses a "kitchen sink" configuration file (without samplers, more about that later). In that test case a `Resource` object is created first, followed by a `TracerProvider` object. The execution will stop in every function that is called to create those objects. Use `c` to move forward. Type the name of any function attribute to check its value.

The current schema has a recursive definition for the `Sampler`s. This doesn't work with this implementation because it is not possible to define the paths for a sampler if the sampler can include infinite other samplers. I think that we don't need a recursive definition of the samplers (I don't think anyone needs to define a sampler that includes an arbitrarily long sequence of nested samplers). I think to solve this situation we can separate the definition of samplers into samplers that can contain _one_ sampler and _terminal_ samplers which are contained by other samplers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@
 # Node.js files for tools (e.g. markdown-toc)
 node_modules/
 package-lock.json
+
+# Python cache
+__pycache__
--- a/prototypes/python/LICENSE
+++ b/prototypes/python/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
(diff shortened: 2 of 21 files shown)


jack-berg:
Closing since this repository is not meant to house implementations of file configuration.

Issue #55, opened 2023-10-30
Default config file template that incorporates standard env vars


tylerbenson:
Consider publishing/maintaining a default template that contains all the standard environment variables being mapped to the file config equivalent.

I think this would be important to enable users with complex env var config setup to successfully migrate to file config.

codeboten:
Thanks for the issue @tylerbenson, i like this idea, we could take https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml as a starting point and use the environment variables that are already documented there.

PR #64, opened 2024-01-05
Define resource.detectors.attributes.included/excluded 

jack-berg:
The specification defines the concept of [resource detectors](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment), which automatically provide resource information based on the environment. 

We need to give the user the ability to disable these keys, as they can contain sensitive information (i.e. `process.command_line` often contains secrets for java) or just not be valuable to the user.

This PR introduces `resource.disabled_attribute_keys`, an array of strings defining attribute keys from resource detectors that should be disabled. Entries in the array can exactly match an attribute key, or can use a basic pattern matching syntax (inherited from metrics view [instrument selection criteria](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#instrument-selection-criteria)) supporting `?` single character any match, and `*` any number of character any match including none.

The effect is that `disabled_attribute_keys` acts as a disallow list. We may wish to add an allow list variant in the future, but disabling keys is much more common so let's start there.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -377,15 +377,23 @@ resource:
     #
     # Environment variable: OTEL_SERVICE_NAME
     service.name: !!str "unknown_service"
+  # Configure resource detectors.
+  detectors:
+    # Configure attributes provided by resource detectors.
+    attributes:
+      # Configure list of attribute key patterns to include from resource detectors. If not set, all attributes are included.
+      #
+      # Attribute keys from resource detectors are evaluated to match as follows:
+      #   * If the value of the attribute key exactly matches.
+      #   * If the value of the attribute key matches the wildcard pattern, where '?' matches any single character and '*' matches any number of characters including none.
+      included:
+        - process.*
+      # Configure list of attribute key patterns to exclude from resource detectors. Applies after .resource.detectors.attributes.included (i.e. excluded has higher priority than included).
+      #
+      # Attribute keys from resource detectors are evaluated to match as follows:
+      #   * If the value of the attribute key exactly matches.
+      #   * If the value of the attribute key matches the wildcard pattern, where '?' matches any single character and '*' matches any number of characters including none.
+      excluded:
+        - process.command_args
   # Configure the resource schema URL.
   schema_url: https://opentelemetry.io/schemas/1.16.0
-  # Configure disabled resource attribute keys provided by resource detectors.
-  #
-  # Attribute keys are evaluated to match disabled_attribute_keys in the following manner:
-  #   * If the value of the attribute key exactly matches.
-  #   * If the value of the attribute key matches the wildcard pattern, where '?' matches any single character and '*' matches any number of characters including none.
-  disabled_attribute_keys:
-    # Configure exact match disabled attribute key.
-    - process.command_args
-    # Configure wildcard match disabled attribute key(s).
-    - process.*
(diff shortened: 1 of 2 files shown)


examples/kitchen-sink.yaml
```diff
@@ -379,3 +379,15 @@ resource:
...
+    # Configure disabled resource attribute keys provided by resource detectors.
+    #
+    # Attribute keys are evaluated to match disabled_attribute_keys in the following manner:
+    #   * If the value of the attribute key exactly matches.
+    #   * If the value of the attribute key matches the wildcard pattern, where '?' matches any single character and '*' matches any number of characters including none.
+    disabled_attribute_keys:
+      # Configure exact match disabled attribute key.
+      - process.command_args
+      # Configure wildcard match disabled attribute key(s).
+      - process.*
```
jack-berg, line 393: Discussed in the 1/8/24 configuration SIG:

- Decided to move this property up to the top level instead of nesting under detectors. This solves any ambiguity issues about whether attributes like `telemetry.sdk.*` are included in the set, since those may be part of the default resource or provided by resource detectors. 
- Discussed whether wildcard syntax is sufficient. Opened #55 to discuss. I propose we stick with wildcard syntax for now and use user feedback to determine if full regex support is needed before a 1.x release.
- Discussed whether it would also be useful to match on resource values. Answer is probably? Could add `disabled_attribute_values` in a separate PR.

examples/kitchen-sink.yaml
```diff
@@ -379,3 +379,13 @@ resource:
     service.name: !!str "unknown_service"
   # Configure the resource schema URL.
   schema_url: https://opentelemetry.io/schemas/1.16.0
+  # Configure disabled resource attribute keys provided by resource detectors.
+  #
+  # Attribute keys are evaluated to match disabled_attribute_keys in the following manner:
+  #   * If the value of the attribute key exactly matches.
+  #   * If the value of the attribute key matches the wildcard pattern, where '?' matches any single character and '*' matches any number of characters including none.
+  disabled_attribute_keys:
```
codeboten, line 387: Maybe this could be an `included`/`excluded` list under `attributes`.

codeboten (commented):
As discussed in the 29-Apr-2024 meeting, the recommendation here is to follow the same pattern for this and prometheus resource attributes in #64

examples/kitchen-sink.yaml
jack-berg: Yeah I like that.

jack-berg: Scratch that. `.resource.attributes` is currently a map with key values which directly reflect the resource attributes. Its odd to have special-case `included` and `excluded` keys which have values following different rules. 

Lets nest `included` and `excluded` under something which indicates that we're referring to attributes from [resource detectors](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment). Maybe something like:
```
resource:
  detectors:
    attributes:
      included:
        - foo*
        - b?r
      excluded:
        - baz
``` 
Nesting `included` and `excluded` under `.resource.detectors.attributes` because:
- We might want to configure other things related to resource detectors
- If `included` and `excluded` are directly under `.resource.detectors`, then its not clear that we're including / excluding attributes vs. detectors themselves

jack-berg: Pushed a commit which reflect this. ^^

codeboten (approved):
👍🏻

PR #67, opened 2024-01-15
[exporter/prometheus] add with_resource_constant_labels

codeboten:
This add support for the following from the spec:

> A Prometheus Exporter MAY offer configuration to add
> resource attributes as metric attributes. By default,
> it MUST NOT add any resource attributes as metric
> attributes. The configuration SHOULD allow the user
> to select which resource attributes to copy (e.g.
> include / exclude or regular expression based). Copied
> Resource attributes MUST NOT be excluded from target_info.

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -134,6 +134,8 @@ meter_provider:
             without_type_suffix: false
             # Configure Prometheus Exporter to produce metrics without a scope info metric.
             without_scope_info: false
+            # Configure Prometheus Exporter to add resource attributes as metrics attributes.
+            with_resource_constant_labels: "service\\.[.]*"
     # Configure a periodic metric reader.
     - periodic:
         # Configure delay interval (in milliseconds) between start of two consecutive exports.
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -135,7 +135,11 @@ meter_provider:
             # Configure Prometheus Exporter to produce metrics without a scope info metric.
             without_scope_info: false
             # Configure Prometheus Exporter to add resource attributes as metrics attributes.
-            with_resource_constant_labels: "service\\.[.]*"
+            with_resource_constant_labels:
+              included:
+                - "service.*"
+              excluded:
+                - "otherattribute"
     # Configure a periodic metric reader.
     - periodic:
         # Configure delay interval (in milliseconds) between start of two consecutive exports.
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -136,10 +136,12 @@ meter_provider:
             without_scope_info: false
             # Configure Prometheus Exporter to add resource attributes as metrics attributes.
             with_resource_constant_labels:
+              # Configure resource attributes to be included, in this example attributes starting with service.
               included:
-                - "service.*"
+                - "service*"
+              # Configure resource attributes to be excluded, in this example attribute service.attr1.
               excluded:
-                - "otherattribute"
+                - "service.attr1"
     # Configure a periodic metric reader.
     - periodic:
         # Configure delay interval (in milliseconds) between start of two consecutive exports.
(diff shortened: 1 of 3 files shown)


schema/meter_provider.json
```diff
@@ -93,6 +93,9 @@
                 },
                 "without_scope_info": {
                     "type": "boolean"
+                },
+                "with_resource_as_constant_labels": {
```
jack-berg, line 97: Ideally this would be named in the spec.. If we merge this can followup with a PR to the spec which proposes the same name. Not sure order matters.

examples/kitchen-sink.yaml
```diff
@@ -125,6 +125,8 @@ meter_provider:
             without_units: false
             without_type_suffix: false
             without_scope_info: false
+            # Configure Prometheus Exporter to add resource attributes as metrics attributes.
+            with_resource_as_constant_labels: "service\\.[.]*"
```
jack-berg, line 129: In #64 I proposed using the view instrument name wildcard syntax, where `?` matches any single character, and `*` matches any number of characters including none. 

The argument being that this avoids having to answer questions about which flavor of regex is used, and whether we can actually accommodate that across languages.

The downside of a simpler wildcard syntax is that you likely need to switch this parameter from a `string` type to an array of strings, since you can easily run into situations where you can't express the desired set in a single expression.

codeboten: Right, another downside is with a wildcard there isn't a mechanism for exclusions built-in

schema/meter_provider.json
asafm: I just finished the PR for this in the Java SDK.
I was thinking more about a system property named `otel.exporter.prometheus.resource.attributes.add.regexp`; hence, can we call this field `add_resource_attributes_regexp`?

codeboten: This is now named in the spec as `with_resource_constant_labels`

examples/kitchen-sink.yaml
jack-berg: What do you think of #71?

jack-berg: The benefit of #71 is that we get ahead of what I think will be a recurring pattern of wanting to provide a way in file config to specify an allow list or disallow list with pattern matching. Regex comes with the complexity of specifying the version of regular expressions, and if a language doesn't support the specification, we'll have worse interoperability because regexes will need to be adjusted for an individual language's requirements. Glob pattern matching is straight forward to implement and gets most of there of what you want to do with regex.

codeboten: What are your thoughts on how one would use #71 to include everything except something that matches? I'm thinking specifically in the collector where someone may want to include all resource attributes except for maybe one of them.

codeboten: I like the simplicity of wildcards better, so long as we retain the ability to have an include or exclude match, then that would solve my use-case

codeboten: Various components in the collector support the following for include/exclude rules:

```yaml
include:
    - /var/log/pods/default_*/*/*.log
exclude: []
```

@jack-berg if you're ok with this, it can be added to #71

jack-berg: Yeah that sounds great. Do you want to reflect that in this PR and merge? Then I can update #71 in a followup to reflect that syntax we use here.

examples/kitchen-sink.yaml
```diff
@@ -134,6 +134,12 @@ meter_provider:
             without_type_suffix: false
             # Configure Prometheus Exporter to produce metrics without a scope info metric.
             without_scope_info: false
+            # Configure Prometheus Exporter to add resource attributes as metrics attributes.
+            with_resource_constant_labels:
+              included:
```
jack-berg, line 139: Can you add comments describing the semantics of these properties? The `kitchen-sink.yaml` is the best we currently have for describing behavioral expectations of implementations.

jack-berg (approved):

examples/kitchen-sink.yaml
codeboten: Updated

PR #69, opened 2024-01-30
Refactor for env vars

jack-berg:
Refactor the schema to be more friendly to an environment variable override scheme derived from the model names.

An example of such a scheme is described in [this document](https://docs.google.com/document/d/1yPfdf6fsxWY7onWU_PLmIIPs14H6pTYbyI7OboiODCw/edit?pli=1#heading=h.r8b09z6u7mwh).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,31 @@ Allowable changes:
 - For major versions: All changes are permitted.
 - For minor versions: TBD
 
-## Naming guidelines
+## Schema Modeling Guidelines
 
-The following defines guidelines used to produce configuration schema:
+The following guidelines are used to model the configuration schema:
 
-1. To remove redundant information from the configuration file, prefixes for data produced by each of the providers
+* To remove redundant information from the configuration file, prefixes for data produced by each of the providers
 will be removed from configuration options. For example, under the `meter_provider` configuration, metric readers will be
 identified by the word `readers` rather than by `metric_readers`. Similarly, the prefix `span_` will be dropped for tracer
-provider configuration, and `logrecord` for logger provider.
+provider configuration, and `logrecord` for logger provider. 
+* Avoid use of object arrays, which are difficult to manipulate with environment variable overrides. Instead, separate the configuration of the elements from
+  their ordering. If multiple of the same element are allowed, use keys of matching `^(?<elementName>[a-zA-Z0-9_]*)(\/(?<elementId>[a-zA-Z0-9_]*))?$`. For
+  example, when configuring the span processor pipeline, use:
+   ```
+   tracer_provider:
+     processors:
+       batch/local:
+         exporter:
+           otlp:
+             endpoint: http://localhost:4317
+       batch/remote:
+         exporter:
+           otlp:
+             endpoint: http://remote-host:4317      
+     processor_pipeline: [batch/local, batch/remote]
+   ```
+* Use keys matching `^[a-zA-Z0-9_]*$`. Keys with other characters are difficult to manipulate with environment variable overrides. If modeling a property which
+  consists of key values but whose keys do not conform to these restrictions, model it as a comma separated string value of the
+  form: `<key1>=<value1>,<key2>=<value2>`. For example, HTTP headers commonly use `-`, which is not allowed, but can be represented
+  with `Api-Key=123,X-Tenant-Id=abc`.
--- a/examples/anchors.yaml
+++ b/examples/anchors.yaml
@@ -8,35 +8,36 @@ exporters:
     certificate: /app/cert.pem
     client_key: /app/cert.pem
     client_certificate: /app/cert.pem
-    headers:
-      api-key: !!str 1234
+    headers: "api-key=1234"
     compression: gzip
     timeout: 10000
 
 logger_provider:
   processors:
-    - batch:
-        exporter:
-          otlp:
-            # expand the otlp-exporter anchor
-            <<: *otlp-exporter
+    batch:
+      exporter:
+        otlp:
+          # expand the otlp-exporter anchor
+          <<: *otlp-exporter
+  processor_pipeline: [batch]
 
 meter_provider:
   readers:
-    - periodic:
-        interval: 5000
-        timeout: 30000
-        exporter:
-          otlp:
-            # expand the otlp-exporter anchor and add metric specific configuration
-            <<: *otlp-exporter
-            temporality_preference: delta
-            default_histogram_aggregation: base2_exponential_bucket_histogram
+    periodic:
+      interval: 5000
+      timeout: 30000
+      exporter:
+        otlp:
+          # expand the otlp-exporter anchor and add metric specific configuration
+          <<: *otlp-exporter
+          temporality_preference: delta
+          default_histogram_aggregation: base2_exponential_bucket_histogram
 
 tracer_provider:
   processors:
-    - batch:
-        exporter:
-          otlp:
-            # expand the otlp-exporter anchor
-            <<: *otlp-exporter
+    batch:
+      exporter:
+        otlp:
+          # expand the otlp-exporter anchor
+          <<: *otlp-exporter
+  processor_pipeline: [batch]
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -29,68 +29,76 @@ attribute_limits:
 # Configure logger provider.
 logger_provider:
   # Configure log record processors.
+  #
+  # Entries not included in logger_provider.processor_pipeline are ignored.
+  #
+  # Entry keys MUST match the expression "^(?<processorName>[a-zA-Z0-9_]*)(\/[a-zA-Z0-9_]*)?$". For example:
+  #  * "batch" - refers to batch log record processor.
+  #  * "batch/collector" - refers to batch log record processor, while disambiguating from other batch log record processors (i.e. "batch/vendorA")
+  #  * "batch^&" - is invalid because it contains characters which are not allowed.
   processors:
     # Configure a batch log record processor.
-    - batch:
-        # Configure delay interval (in milliseconds) between two consecutive exports.
-        #
-        # Environment variable: OTEL_BLRP_SCHEDULE_DELAY
-        schedule_delay: 5000
-        # Configure maximum allowed time (in milliseconds) to export data.
-        #
-        # Environment variable: OTEL_BLRP_EXPORT_TIMEOUT
-        export_timeout: 30000
-        # Configure maximum queue size.
-        #
-        # Environment variable: OTEL_BLRP_MAX_QUEUE_SIZE
-        max_queue_size: 2048
-        # Configure maximum batch size.
-        #
-        # Environment variable: OTEL_BLRP_MAX_EXPORT_BATCH_SIZE
-        max_export_batch_size: 512
-        # Configure exporter.
-        #
-        # Environment variable: OTEL_LOGS_EXPORTER
-        exporter:
-          # Configure exporter to be OTLP.
-          otlp:
-            # Configure protocol.
-            #
-            # Environment variable: OTEL_EXPORTER_OTLP_PROTOCOL, OTEL_EXPORTER_OTLP_LOGS_PROTOCOL
-            protocol: http/protobuf
-            # Configure endpoint.
-            #
-            # Environment variable: OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_LOGS_ENDPOINT
-            endpoint: http://localhost:4318
-            # Configure certificate.
-            #
-            # Environment variable: OTEL_EXPORTER_OTLP_CERTIFICATE, OTEL_EXPORTER_OTLP_LOGS_CERTIFICATE
-            certificate: /app/cert.pem
-            # Configure mTLS private client key.
-            #
-            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_KEY, OTEL_EXPORTER_OTLP_LOGS_CLIENT_KEY
-            client_key: /app/cert.pem
-            # Configure mTLS client certificate.
-            #
-            # Environment variable: OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE, OTEL_EXPORTER_OTLP_LOGS_CLIENT_CERTIFICATE
-            client_certificate: /app/cert.pem
-            # Configure headers.
-            #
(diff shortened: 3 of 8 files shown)


marcalff (commented):
If I understand correctly, the motivation for replacing the array under the tracer provider span processors is that accessing arrays by index causes difficulties.

This is replaced by processor_pipeline, which is ... an array.

Why would accessing elements in the processor_pipeline array be any easier compared to accessing elements in the processor array ?

What is gained by this model change ?

For example, if during a merge a user wants to add one more processor to send traces to the console for debug, without knowing what processors are already defined, why would the syntax to add a simple/console entry under the pipeline be better compared to the syntax to add a new simple/exporter/console processor under processors ?

jack-berg:
Arrays of strings are more manageable than arrays of objects.

If you have array of strings, you can say that you can override it via an environment variable which is comma separated. In the example of adding a new console entry, you might have:
```
tracer_provider:
  processors:
    batch:
      exporter:
        otlp:
  processor_pipeline: [batch]
```
And set the following environment variables:

```
OTEL_CONF_TRACER_PROVIDER__PROCESSOR_PIPELINE=batch,simple
OTEL_CONF_TRACER_PROVIDER__PROCESSORS__SIMPLE__EXPORTER__CONSOLE={}
```
Which would result in:
```
tracer_provider:
  processors:
    batch:
      exporter:
        otlp:
    simple:
      simple:
        console:
  processor_pipeline: [batch,simple]
```
This assumes the rules are such that if you have an environment variable that references a structure that does not exist, you create that structure. I.e. `OTEL_CONF_TRACER_PROVIDER__PROCESSORS__SIMPLE__EXPORTER__CONSOLE` references `.tracer_provider.processors.simple.exporter.console`, so we create that structure. 

Supporting arrays of objects requires that you support a more complicated env var syntax discussed, which has the types of sharp edges discussed in this [document](https://docs.google.com/document/d/1yPfdf6fsxWY7onWU_PLmIIPs14H6pTYbyI7OboiODCw/edit#heading=h.r8b09z6u7mwh).

marcalff:
> Supporting arrays of objects requires that you support a more complicated env var syntax discussed, which has the types of sharp edges discussed in this [document](https://docs.google.com/document/d/1yPfdf6fsxWY7onWU_PLmIIPs14H6pTYbyI7OboiODCw/edit#heading=h.r8b09z6u7mwh).

Yes, I have seen this document.

Have you seen this https://github.com/open-telemetry/opentelemetry-specification/issues/3752#issuecomment-1925944612 yet ?

Instead of several environment variables like:

* `export OTEL_CONF_PATH_TO_SOME_NODE = "value"`
* `export OTEL_CONF_PATH_TO_SOME_OTHER_MODE = "other value"`

I am proposing to use some existing tool (`yq`), and put all the logic in a yq script, as in:

* `export OTEL_CONF_MERGE = '.path.to.some.node = "value" | .path.to.some.other.node = "other value"'`

Because the path to the node appears on the right hand side, there is no longer a need to encode a yaml path as an environment variable name, taking care of some rough edges.

Because several changes can be included in the same script using a pipe, the entire description for, say, a complex trace sampler, can fit into a single environment variable instead of several related env variables (which looks better to me).

Instead of bending the schema to accommodate the tooling restrictions, can we improve the tooling instead ?

For headers for example, my understanding from this PR is that headers are a single string now. I assume it would contain something encoded, like a key/value comma delimited string ? The previous schema was better for this area, IMO.

README.md
```diff
@@ -31,11 +31,31 @@ Allowable changes:
...
+     processors:
+       batch/local:
+         exporter:
+           otlp:
+             endpoint: http://localhost:4317
+       batch/remote:
+         exporter:
+           otlp:
+             endpoint: http://remote-host:4317      
+     processor_pipeline: [batch/local, batch/remote]
```
trask, line 56: why is `processor_pipeline` needed?

examples/kitchen-sink.yaml
```diff
@@ -12,358 +12,404 @@ file_format: "0.1"
...
-# Environment variable: OTEL_SDK_DISABLED
+# Environment variable: OTEL_CONF_DISABLED
 disabled: false
 
 # Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
 attribute_limits:
   # Configure max attribute value size.
   #
-  # Environment variable: OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT
+  # Environment variable: OTEL_CONF_ATTRIBUTE_LIMITS__ATTRIBUTE_VALUE_LENGTH_LIMIT
```
trask, line 22: do you think the `_CONF_` portion is necessary?

README.md
jack-berg: Versus just using relying on the map of `.tracer_provider.processors`? 

Order matters with processors and there's no guarantee that libraries used to parse YAML will provide access to map entries in the order they appear in the YAML. In java, I believe snakeYAML uses `LinkedHashMap` internally to represent objects so that order is retained, but there's no guarantee this is true in other implementation.

Additionally, the process pipeline gives an easy way to disable a processor via env variable without having to support some special env var syntax to "erase" an object. I.e. to remove `batch/remote`, set `OTEL_CONF_TRACER_PROVIDER__PROCESSOR_PIPELINE=batch/local`.

examples/kitchen-sink.yaml
jack-berg: It seemed like a good way to clearly distinguish between this new environment variable schema and the old. Not strictly required, but the common prefix catches the eye and there is no mistaking which scheme it refers to.

We of course should weigh this against characters it adds to every single environment variable.

examples/kitchen-sink.yaml
```diff
@@ -12,358 +12,404 @@ file_format: "0.1"
...
-            # Environment variable: OTEL_EXPORTER_OTLP_TIMEOUT, OTEL_EXPORTER_OTLP_TRACES_TIMEOUT
-            timeout: 10000
-            # Configure client transport security for the exporter's connection.
-            #
-            # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_TRACES_INSECURE
-            insecure: false
+    batch:
+      # Configure delay interval (in milliseconds) between two consecutive exports.
+      #
+      # Environment variable: OTEL_CONF_TRACER_PROVIDER__PROCESSORS__BATCH__SCHEDULE_DELAY
```
trask, line 298: wdyt of shortening the top-level config keys:
* `tracer_provider` -> `traces`
* `meter_provider` -> `metrics`
* `log_provider` -> `logs`

or if we think that's too ambiguous (e.g. tracing instrumentation could be considered part of `traces`), maybe having a top-level `sdk` config node, then the resulting env vars would be something like:
* `sdk_traces_`
* `sdk_metrics_`
* `sdk_logs_`

examples/kitchen-sink.yaml
trask: I'm personally in favor of dropping it if the plan is for these to be the (one set of) recommended env vars going forward

examples/anchors.yaml
```diff
@@ -8,35 +8,36 @@ exporters:
     certificate: /app/cert.pem
     client_key: /app/cert.pem
     client_certificate: /app/cert.pem
-    headers:
-      api-key: !!str 1234
+    headers: "api-key=1234"
```
marcalff, line 11: I have some concerns with this.

Will it be possible later to not have the value appear in the config.yaml / environment variables content, and have a yaml node that represents a secret instead ?

examples/kitchen-sink.yaml
```diff
@@ -12,358 +12,404 @@ file_format: "0.1"
...
+      exporter:
+        # Configure exporter to be OTLP.
+        otlp:
+          # Configure protocol.
+          #
+          # Environment variable: OTEL_CONF_LOGGER_PROVIDER__PROCESSORS__BATCH__EXPORTER__OTLP__PROTOCOL
+          protocol: http/protobuf
+          # Configure endpoint.
+          #
+          # Environment variable: OTEL_CONF_LOGGER_PROVIDER__PROCESSORS__BATCH__EXPORTER__OTLP__ENDPOINT
```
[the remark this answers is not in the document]
jack-berg: > I understand the ambiguity that brings tho...

To make this concrete.. 

Suppose we want to change the `max_queue_size` of the batch span processor given the following config
```
tracer_provider:
  processors:
    batch:
      exporter:
        otlp:
          endpoint: http://endpoint1
  processor_pipeline: [batch]
```

If we use spring's mappings, we'd set something like:
```
OTEL_CONF_TRACERPROVIDER_PROCESSORS_BATCH_MAXQUEUESIZE=10
```

In order to apply this override, we have to know the structure of the schema ahead of time. If we don't, and just blindly try to update the raw YAML, we'd get:
```
tracer_provider:
  processors:
    batch:
      exporter:
        otlp:
          endpoint: http://endpoint1
  processor_pipeline: [batch]
tracerprovider:
  processors:
    batch:
      maxqueuesize: 10
```
In this case, we know that the batch processor has a property called `max_queue_size`. So when evaluating, we can take the position of this property `.tracer_provider.processors.batch.max_queue_size`, apply a generic transformation yielding `OTEL_CONF_TRACERPROVIDER_PROCESSORS_BATCH_MAXQUEUESIZE` and check if any environment variable is set which matches.

But we don't know the schema ahead of time for custom components. Consider a custom span exporter named `foo` with a property called `my_property`:
```
tracer_provider:
  processors:
    batch:
      exporter:
        foo:
          my_property: http://endpoint1
```
Because we don't know the schema of the foo exporter ahead of time, we have to way to tell that `OTEL_TRACERPROVIDER_PROCESSORS_BATCH_EXPORTER_FOO_MYPROPERTY` should target `my_property` or `myproperty`. 


Prior art that uses the `__` include python's [dynaconf](https://dynaconf.readthedocs.io/en/docs_223/guides/environment_variables.html#nested-keys-in-dictionaries-via-environment-variables) and [.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-8.0#non-prefixed-environment-variables).

examples/kitchen-sink.yaml
jack-berg: The competing concerns with shortening top level keys are that on one hand we want shorter env vars, but on the other hand we want the config to mirror the SDK concepts of the specification. I.e. renaming `tracer_provider` to `traces` blurs the fact that you're configuring a tracer provider. 

> or if we think that's too ambiguous (e.g. tracing instrumentation could be considered part of traces), maybe having a top-level sdk config node, then the resulting env vars would be something like:

There's other to level attributes as well like `.attribute_limits`, `.disabled`, and `.resource`. The `sdk` disambiguates, but becomes repetitive when applied to all the properties. Begs the question of why not just nest all the top level properties under `.sdk` (i.e. `.sdk.tracer_provider`, `.sdk.meter_provider`, or `.sdk.traces`, `.sdk.metrics`)

The instrumentation question is interesting because its not clear to what extent instrumentation config will be normalized across languages, and to what extent we should even try this (if the long term vision is instrumentation moving to native, we can't force libraries to follow standard configuration).

One answer is to leave all the SDK config top level, and say that all instrumentation config is nested under `.instrumentation`, or something to that affect.

examples/kitchen-sink.yaml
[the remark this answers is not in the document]
trask: > Prior art that uses the __ include python's [dynaconf](https://dynaconf.readthedocs.io/en/docs_223/guides/environment_variables.html#nested-keys-in-dictionaries-via-environment-variables) and [.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-8.0#non-prefixed-environment-variables).

👍

examples/anchors.yaml
jack-berg: Sorry I'm not grasping the question. Can you elaborate?

marcalff: When some fields are secrets, like a password, or here an auth token:

```
headers:
  foo: 12
  bar: 24
  api-key: actual_auth_token <-- visible in plain view
```

The model can be changed to separate key names and key values:

```
headers:
  - key:
    name: foo
    value: 12
  - key:
    name: bar
    value: 24
  - key:
    name: api-key
    value: actual_auth_token <-- still visible in plain view
```

and then, a key value can be represented by an alternative node, which is a named proxy:

```
headers:
  - key:
    name: foo
    value: 12
  - key:
    name: bar
    value: 24
  - key:
    name: api-key
    secret:
      secret_name: "vendor_auth_token"
```

The secret name, "vendor_auth_token", is just a name and not the actual value.
The actual value (actual_auth_token) is retrieved from a secure location, referenced by name (vendor_auth_token).

The concern is that serializing key/value pairs in a comma delimited string prevents this.

jack-berg: Ah I see. And that syntax is what is used in kubernetes, so there's definitely (popular) prior art for this type of syntax.

A counter argument is that `OTEL_EXPORTER_OTLP_HEADERS` has been around for a while and I haven't heard many (any?) issues with the fact that its a comma separated list of key value pairs. This design is essentially saying that because a comma separated list was "good enough" for `OTEL_EXPORTER_OTLP_HEADERS`, that its probably good enough for file configuration too. The trade off of course is expressiveness vs. ease of overriding w/ environment variables. The syntax you give above is highly expressive but not easily overridden with environment variables. Treating headers as a flattened comma separated list is not as expressive but is rally easy to override with environment variables for _most_ cases.

marcalff: In a world pre `config.yaml`, the application is configuring the SDK by writing explicit code.

When doing so, some configuration can be done using environment variables as a help, or ignoring environment variables and doing everything programmatically, or both.

Disclosure: For a major app at work, not a single environment variable is used, everything is done programmatically, finding the configuration data "elsewhere".

Now, in a setup using `config.yaml`, the main advantage of unifying all the configuration in a yaml document is that the entire configuration can change, without having to change a single line of code in the application.

Having auth tokens in a key/value delimited string in an environment variable (for people willing to use this), or in code (for people not willing to use this) was workable before, because the app code was ad hoc anyway.

In a world where all is unified with a yaml config file, parts which stands out for specific reasons (as in, I would like all the config in yaml ... **_except_** the sensitive part) are more likely to cause friction and stand out, compared to a pre config.yaml usage. What was good enough before may not be good enough now.

If there is a choice to make between config.yaml and security, it is already made: config.yaml will not be used, period.

So, all the work with configuration in general should have provisions to avoid these pitfalls.

As a side note, using key/value pairs in comma delimited strings (this PR) also broke the use case involving variables substitution:

```
headers:
  foo: 12
  bar: 34
  api-key: ${MY_VENDOR_TOKEN}
```

This setup is no longer working.

PR #76, opened 2024-03-18
Add sdk-config.yaml starter template w/ references to env vars

### LINKED ISSUE #55: Default config file template that incorporates standard env vars

jack-berg:
Resolves #55.

Part of the configuration working group recommendation on how to move forward with [#3752](https://github.com/open-telemetry/opentelemetry-specification/issues/3752) as [described here](https://github.com/open-telemetry/opentelemetry-specification/issues/3752#issuecomment-1995582317).

- Part of https://github.com/open-telemetry/opentelemetry-specification/pull/3948
- Adds a new `sdk-migration-config.yaml` to the examples directory, which includes env var substitution references to all env vars which map cleanly to file config
- Adds a new `sdk-config.yaml` to the example directory, which is identical to `sdk-migration-config.yaml`, but without env var substitution references
- Adds a new step to the makefile build which performs env var substitution using [js envsub](https://github.com/danday74/envsub) which allows validation to take place after substitution
- Updates the schema so all types can be null. This ensures a config file is still valid after performing env var substitution with an undefined env var.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,15 @@ This repository contains the JSON schema that defines the OpenTelemetry configur
 - code generation
 - broad support across languages
 
+## Starter templates
+
+The [examples](./examples) repository contains a variety of sample configuration files to help get started and illustrate useful patterns. The following are noteworthy:
+
+- [sdk-migration-config.yaml](./examples/sdk-migration-config.yaml): Includes env var substitution references to all [standard environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md) which map cleanly to file configuration (see notes in the example for the set of env vars which are not referenced). Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][]. This is a great starting point for transitioning from env var based configuration to file based configuration.
+- [sdk-config.yaml](./examples/sdk-config.yaml): Represents the typical default configuration. This is a good starting point if you are not using env var based configuration or wish to transition fully to file baed configuration. Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][].
+
+[env var substitution]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution
+
 ## Code generation
 
 There are [several tools](https://json-schema.org/implementations.html) available to generate code from a JSON schema. The following shows an example for generating code from the JSON schema in Go:
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This repository contains the JSON schema that defines the OpenTelemetry configur
 The [examples](./examples) repository contains a variety of sample configuration files to help get started and illustrate useful patterns. The following are noteworthy:
 
 - [sdk-migration-config.yaml](./examples/sdk-migration-config.yaml): Includes env var substitution references to all [standard environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md) which map cleanly to file configuration (see notes in the example for the set of env vars which are not referenced). Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][]. This is a great starting point for transitioning from env var based configuration to file based configuration.
-- [sdk-config.yaml](./examples/sdk-config.yaml): Represents the typical default configuration. This is a good starting point if you are not using env var based configuration or wish to transition fully to file baed configuration. Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][].
+- [sdk-config.yaml](./examples/sdk-config.yaml): Represents the typical default configuration. This is a good starting point if you are not using env var based configuration or wish to transition fully to file based configuration. Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][].
 
 [env var substitution]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution
 
--- a/examples/sdk-config.yaml
+++ b/examples/sdk-config.yaml
@@ -50,7 +50,7 @@ tracer_provider:
             # Configure protocol.
             protocol: http/protobuf
             # Configure endpoint.
-            endpoint: http://localhost:-4318
+            endpoint: http://localhost:4318
             # Configure certificate.
             certificate:
             # Configure mTLS private client key.
@@ -117,7 +117,7 @@ meter_provider:
             # Configure protocol.
             protocol: http/protobuf
             # Configure endpoint.
-            endpoint: http://localhost:-4318
+            endpoint: http://localhost:4318
             # Configure certificate.
             certificate:
             # Configure mTLS private client key.
--- a/examples/sdk-config.yaml
+++ b/examples/sdk-config.yaml
@@ -1,0 +1,206 @@
+# sdk-config.yaml is a typical starting point for configuring the SDK, including exporting to
+# localhost via OTLP.
+
+# NOTE: With the exception of env var substitution syntax, SDKs ignore environment variables
+# when interpreting config files.
+#
+# This template includes env var substitution references (i.e. ${MY_ENV}) for all spec defined
+# env vars (https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)
+# which map cleanly to file configuration. For example, OTEL_SDK_DISABLED is referenced below,
+# but OTEL_TRACES_EXPORTER is not since it does not map well to the hierarchical structure of
+# file configuration.
+#
+# Because env vars are ignored except for env var substitution, if "disabled: ${OTEL_SDK_DISABLED:-false}"
+# is replaced with "disabled: false", setting OTEL_SDK_DISALBED env var will not have any effect.
+# See https://opentelemetry.io/docs/specs/otel/configuration/file-configuration/ for more information.
+# The following spec defined env vars are NOT referenced and are thus ignored:
+#
+# - OTEL_RESOURCE_ATTRIBUTES
+# - OTEL_LOG_LEVEL
+# - OTEL_PROPAGATORS
+# - OTEL_TRACES_SAMPLER
+# - OTEL_TRACES_SAMPLER_ARG
+# - OTEL_EXPORTER_ZIPKIN_ENDPOINT
+# - OTEL_EXPORTER_ZIPKIN_TIMEOUT
+# - OTEL_EXPORTER_PROMETHEUS_HOST
+# - OTEL_EXPORTER_PROMETHEUS_PORT
+# - OTEL_TRACES_EXPORTER
+# - OTEL_METRICS_EXPORTER
+# - OTEL_LOGS_EXPORTER
+# - OTEL_METRICS_EXEMPLAR_FILTER
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_ENDPOINT
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_INSECURE
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_CERTIFICATE
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_CLIENT_KEY
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_CLIENT_CERTIFICATE
+# - OTEL_EXPORTER_OTLP_HEADERS, OTEL_EXPORTER_OTLP_{SIGNAL}_HEADERS
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_COMPRESSION
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_TIMEOUT
+# - OTEL_EXPORTER_OTLP_{SIGNAL}_PROTOCOL
+
+# The file format version
+file_format: "0.1"
+
+# Configure if the SDK is disabled or not.
+disabled: ${OTEL_SDK_DISABLED:-false}
+
+# Configure resource for all signals.
+resource:
+  # Configure resource attributes.
+  attributes:
+    # Configure `service.name` resource attribute
+    service.name: ${OTEL_SERVICE_NAME:-unknown_service}
+
+# Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
+attribute_limits:
+  # Configure max attribute value size.
+  attribute_value_length_limit: ${OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT}
+  # Configure max attribute count.
+  attribute_count_limit: ${OTEL_ATTRIBUTE_COUNT_LIMIT:-128}
+
+# Configure text map context propagators.
+propagator:
+  composite: [tracecontext, baggage]
+
+# Configure tracer provider.
+tracer_provider:
+  # Configure span processors.
+  processors:
+    # Configure a batch span processor.
+    - batch:
+        # Configure delay interval (in milliseconds) between two consecutive exports.
+        schedule_delay: ${OTEL_BSP_SCHEDULE_DELAY:-5000}
+        # Configure maximum allowed time (in milliseconds) to export data.
+        export_timeout: ${OTEL_BSP_EXPORT_TIMEOUT:-30000}
+        # Configure maximum queue size.
+        max_queue_size: ${OTEL_BSP_MAX_QUEUE_SIZE:-2048}
+        # Configure maximum batch size.
+        max_export_batch_size: ${OTEL_BSP_MAX_EXPORT_BATCH_SIZE:-512}
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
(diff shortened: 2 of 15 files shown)


examples/sdk-config.yaml
```diff
@@ -0,0 +1,206 @@
...
+        # Configure maximum batch size.
+        max_export_batch_size: ${OTEL_BSP_MAX_EXPORT_BATCH_SIZE:-512}
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
+            # Configure endpoint.
+            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:-4318}
```
marcalff, line 86:
```suggestion
            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:4318/v1/traces}
```

examples/sdk-config.yaml
```diff
@@ -0,0 +1,206 @@
...
+        # Configure maximum allowed time (in milliseconds) to export data.
+        timeout: ${OTEL_METRIC_EXPORT_TIMEOUT:-30000}
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
+            # Configure endpoint.
+            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:-4318}
```
marcalff, line 153:
```suggestion
            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:4318/v1/metrics}
```

examples/sdk-config.yaml
```diff
@@ -0,0 +1,206 @@
...
+        # Configure maximum batch size.
+        max_export_batch_size: ${OTEL_BLRP_MAX_EXPORT_BATCH_SIZE:-512}
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
+            # Configure endpoint.
+            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:-4318}
```
marcalff, line 190:
```suggestion
            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:4318/v1/logs}
```

marcalff (approved):
LGTM, with fixes on endpoints.

examples/sdk-config.yaml
tsloughter: That would make it `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`

Also, its currently using `OTEL_EXPORTER_OTLP_PROTOCOL` when I'm guessing it wants either `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` or to actually put the path outside the environment variable default to apply to both:

```
endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/traces
```

marcalff: Good point on the TRACES part. So, something like this then:

```suggestion
            endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/v1/traces}
```

This also means using the TRACES env var for PROTOCOL as well.

jack-berg: We have to decide between the generic version of these properties and the signal specific. I think the generic version of these is likely to be more popular. But the generic version of the property [doesn't require you to configure](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) a path. 

One thing which is still ambiguous is whether the endpoints specified in file config have paths appended automatically if the protocol is `http/protobuf` and the path is omitted. I lean towards yes.

jack-berg:
This is a loose end from https://github.com/open-telemetry/opentelemetry-specification/pull/3948

I've pushed an additional commit and marked as ready for review. Please review when you have the chance.

cc @open-telemetry/configuration-maintainers

examples/sdk-config.yaml
```diff
@@ -0,0 +1,173 @@
...
+        # Configure maximum batch size.
+        max_export_batch_size: 512
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            protocol: http/protobuf
+            # Configure endpoint.
+            endpoint: http://localhost:-4318
```
codeboten, line 53:
```suggestion
            endpoint: http://localhost:4318
```

examples/sdk-config.yaml
```diff
@@ -0,0 +1,173 @@
...
+        # Configure maximum allowed time (in milliseconds) to export data.
+        timeout: 30000
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be OTLP.
+          otlp:
+            # Configure protocol.
+            protocol: http/protobuf
+            # Configure endpoint.
+            endpoint: http://localhost:-4318
```
codeboten, line 120:
```suggestion
            endpoint: http://localhost:4318
```

jack-berg:
@tsloughter, @marcalff, @codeboten can you take another look? Thanks

examples/sdk-config.yaml
codeboten: I'm ok with going with the generic endpoint as an example for now

codeboten (approved):

examples/sdk-config.yaml
marcalff: In my understanding, the only viable setting is:

```
endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/v1/traces}
```

If any of the following are used instead:

```
endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/}
```

```
endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318/}
```

then please point to the spec that mandates to add `v1/traces` in the path, and I don't mean the definition of existing pre config.yaml environment variable `OTEL_EXPORTER_OTLP_ENDPOINT` here, because this must also apply to any variable, including `MY_ENDPOINT`.

examples/sdk-migration-config.yaml
```diff
@@ -0,0 +1,206 @@
...
+# env vars (https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)
+# which map cleanly to file configuration. For example, OTEL_SDK_DISABLED is referenced below,
+# but OTEL_TRACES_EXPORTER is not since it does not map well to the hierarchical structure of
+# file configuration.
+#
+# NOTE: With the exception of env var substitution syntax, SDKs ignore environment variables
+# when interpreting config files.
+#
+# Because env vars are ignored except for env var substitution, if "disabled: ${OTEL_SDK_DISABLED:-false}"
+# is replaced with "disabled: false", setting OTEL_SDK_DISALBED env var will not have any effect.
```
marcalff, line 14: This is circular logic.

> If "disabled: ${OTEL_SDK_DISABLED:-false}" is replaced with "disabled: false"

means that environment variable `OTEL_SDK_DISABLED` is not set, by definition.

Setting `OTEL_SDK_DISABLED` **will have some effect**, because it invalidates the conditions leading to the use of the default value.

Also, there is a typo on `OTEL_SDK_DISALBED`.

marcalff (commented):
Please see more comments.

marcalff (changes requested):
See comments on endpoints.

examples/sdk-config.yaml
tsloughter: I think jack was saying it should just be `endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}`. Which is correct. `v1/traces` is added by default when using the non-signal specific endpoint configuration.

I was only suggesting the signal specific var because you were suggesting the addition of `v1/traces`.

marcalff: To clarify, this is how opentelemetry-cpp parses environment variables (before config.yaml)

```
std::string GetOtlpDefaultHttpTracesEndpoint()
{
  constexpr char kSignalEnv[]  = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT";
  constexpr char kGenericEnv[] = "OTEL_EXPORTER_OTLP_ENDPOINT";
  constexpr char kDefault[]    = "http://localhost:4318/v1/traces";

  std::string value;
  bool exists;

  exists = sdk_common::GetStringEnvironmentVariable(kSignalEnv, value);
  if (exists)
  {
    return value;
  }

  exists = sdk_common::GetStringEnvironmentVariable(kGenericEnv, value);
  if (exists)
  {
    value += "/v1/traces";
    return value;
  }

  return kDefault;
}
```

The `v1/traces` is added when the variable name is `OTEL_EXPORTER_OTLP_ENDPOINT`, it is not added later in the SDK when a trace exporter needs an endpoint.

tsloughter: Argh, sorry, I misunderstood the context from the beginning. This is `endpoint` within the trace provider. So you are right, it should be `endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/traces`

tsloughter: Means the example config won't support setting the path, but it can't support everything?

marcalff: I think there is more to it.

Options are correlated, the path is only used for OTLP HTTP but not for GRPC if I recall correctly, so something like:

```
endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/traces
```

really assumes OTLP HTTP.

Where:

```
endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/v1/traces}
```

can be replaced by the GRPC endpoint (using port 4317 while at it)


The whole point of environment variables is to change parts while not having to touch the config.yaml file.

If the endpoint changes between OTLP HTTP and OTLP GRPC, I don't see how OTEL_EXPORTER_OTLP_ENDPOINT can be used anywhere.

marcalff: So summarize my comments, the content of the config.yaml as written in this PR:

```
            # Configure protocol.
            protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
            # Configure endpoint.
            endpoint: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http://localhost:-4318}
```

does not allow to connect using the OTLP HTTP protocol to an opentelemetry-collector.

tsloughter: Right, it should definitely not be that because it will mean the trace exporter won't send to `v1/traces`.

examples/sdk-migration-config.yaml
jack-berg: Yes, upon rereading this phrasing is definitely confusing. I've reworded the phrasing.

examples/sdk-config.yaml
jack-berg: These is a good convo. Thanks for calling this out @marcalff. 

>> v1/traces is added by default when using the non-signal specific endpoint configuration
> Where, by whom ?

When the `OTEL_EXPORTER_OTLP_ENDPOINT` env var is set, the "where" & "whom" is whatever component is interpreting the env vars and setting up the exporter. According this [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) (as you note) the code is responsible for doing different things with the endpoint value based on whether the signal-specific or generic env var is used, and based on whether the protocol is `grpc` or `http/protobuf`. 

You've suggested the following alternatives:

```
endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/v1/traces}
endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/traces
```

The problem with both of these (and not necessarily a dealbreaker) is that if the sets `OTEL_EXPORTER_OTLP_PROTOCOL=grpc`, then the endpoint breaks. They need to configure `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` in lockstep with `OTEL_EXPORTER_OTLP_PROTOCOL`.

Another alternative is to say that code interpreting the `*.otlp.protocol` and `*.otlp.endpoint` values of the model has additional responsibilities besides plugging these raw values into the exporter. Essentially carrying over the logic interpreting `OTEL_EXPORTER_OTLP_ENDPOINT` to comments:
```
tracer_provider:
  processors:
    - batch:
       exporter:
         otlp:
           # Configure protocol.
           protocol: ${OTEL_EXPORTER_OTLP_PROTOCOL:-http/protobuf}
           # Configure endpoint.
           #
           # When .protocol is set to http/protobuf and the endpoint URL contains no path element, implementations MUST set the path to be "/v1/traces".
           endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}
```

The problem with this is that it makes it impossible to specify that data should be sent do an endpoint without any path at all, and the env var spec seems to have explicitly [carved out an exception for this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp):

> For the per-signal variables (OTEL_EXPORTER_OTLP_<signal>_ENDPOINT), the URL MUST be used as-is without any modification. The only exception is that if an URL contains no path part, the root path / MUST be used (see [Example 2](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#example-2)).


My conclusion is that @marcalff is correct and the only correct alternatives are:

```
endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/v1/traces}
endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/traces
```

And I'm now leaning towards the first options because the idea of appending a path to the `OTEL_EXPORTER_OTLP_ENDPOINT` env var is a code smell to me.
```
endpoint: ${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT:-http://localhost:4318/v1/traces}
```

What do you all think?

tsloughter: Agreed we can't require the code interpreting `endpoint` to append a path. It wouldn't just mean you can't have an empty path, but it would mean you can't have any alternative path at all because it would always append `v1/traces`. There are already vendors using different paths.

I also don't think the second option is a code smell *but* I do agree that because it can't take into account the protocol that the first option is the only real answer.

marcalff: > There are already vendors using different paths.

I can confirm this.

See for example:

https://docs.oracle.com/en-us/iaas/application-performance-monitoring/doc/configure-open-source-tracing-systems.html#GUID-4D941163-F357-4839-8B06-688876D4C61F


    For traces authenticated with public data key, add the following:

    https://<dataUploadEndpoint>/20200101/opentelemetry/public/v1/traces
    For traces authenticated with private data key, add the following:

    https://<dataUploadEndpoint>/20200101/opentelemetry/private/v1/traces

tsloughter: Those still end in `v1/traces`. So having that appended to each would still work fine. I believe some vendors don't have the `v1/traces` at the end.

jack-berg:
@marcalff would appreciate you taking another look when you can. Thanks!

README.md
```diff
@@ -8,6 +8,15 @@ This repository contains the JSON schema that defines the OpenTelemetry configur
...
 - broad support across languages
 
+## Starter templates
+
+The [examples](./examples) repository contains a variety of sample configuration files to help get started and illustrate useful patterns. The following are noteworthy:
+
+- [sdk-migration-config.yaml](./examples/sdk-migration-config.yaml): Includes env var substitution references to all [standard environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md) which map cleanly to file configuration (see notes in the example for the set of env vars which are not referenced). Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][]. This is a great starting point for transitioning from env var based configuration to file based configuration.
+- [sdk-config.yaml](./examples/sdk-config.yaml): Represents the typical default configuration. This is a good starting point if you are not using env var based configuration or wish to transition fully to file baed configuration. Note, SDKs parsing configuration files ignore all env vars besides those referenced via [env var substitution][].
```
marcalff, line 16: Typo: file `baed` configuration

examples/sdk-config.yaml
```diff
@@ -0,0 +1,179 @@
+# sdk-config.yaml is a typical starting point for configuring the SDK, including exporting to
+# localhost via OTLP.
+
+# NOTE: With the exception of env var substitution syntax (i.e. ${MY_ENV}) , SDKs ignore
```
marcalff, line 4: typeset: extra space before comma `(i.e. ${MY_ENV}) ,`

marcalff (approved):
LGTM, see editorial changes.

PR #84, opened 2024-04-25
add console definition for log record exporter

codeboten:
This is consistent with the other signals.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -91,6 +91,12 @@ logger_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
             insecure: false
+    # Configure a simple span processor.
+    - simple:
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be console.
+          console: {}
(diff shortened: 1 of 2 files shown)


examples/kitchen-sink.yaml
```diff
@@ -91,6 +91,12 @@ logger_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
             insecure: false
+    # Configure a simple span processor.
+    - simple:
+        # Configure exporter.
+        exporter:
+          # Configure exporter to be console.
+          console: {}
```
jack-berg, line 99: Its consistent but the console exporter for logs can be tricky to avoid circular loops: 
Log framework -> Otel Log Bridge API -> Otel Log SDK -> Console Exporter -> Log Framework -> 

Maybe we don't need to worry about that for the sake of config. Its tricky to get right, but still possible to configure.

jack-berg (approved):

examples/kitchen-sink.yaml
codeboten: Right, to be clear this was possible to configure even before this change :) we just didnt have a `console` section in the log record exporter called out specifically

PR #90, opened 2024-05-17
Add metric producers to meter_provider configuration

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/4015

Metric producers are registered to metric readers, so i've nested it under the periodic and manual readers.

@gouthamve @jack-berg

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -142,10 +142,10 @@ meter_provider:
               # Configure resource attributes to be excluded, in this example attribute service.attr1.
               excluded:
                 - "service.attr1"
-        # Configure metric producers.
-        producers:
-          # Configure metric producer to be opencensus
-          - opencensus: {}
+      # Configure metric producers.
+      producers:
+        # Configure metric producer to be opencensus
+        - opencensus: {}
     # Configure a periodic metric reader.
     - periodic:
         # Configure delay interval (in milliseconds) between start of two consecutive exports.
@@ -207,10 +207,10 @@ meter_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION
             default_histogram_aggregation: base2_exponential_bucket_histogram
-        # Configure metric producers.
-        producers:
-          # Configure metric producer to be prometheus
-          - prometheus: {}
+      # Configure metric producers.
+      producers:
+        # Configure metric producer to be prometheus
+        - prometheus: {}
     # Configure a periodic metric reader.
     - periodic:
         # Configure exporter.
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -142,6 +142,10 @@ meter_provider:
               # Configure resource attributes to be excluded, in this example attribute service.attr1.
               excluded:
                 - "service.attr1"
+        # Configure metric producers.
+        producers:
+          # Configure metric producer to be opencensus
+          - opencensus: {}
     # Configure a periodic metric reader.
     - periodic:
         # Configure delay interval (in milliseconds) between start of two consecutive exports.
@@ -203,6 +207,10 @@ meter_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION
             default_histogram_aggregation: base2_exponential_bucket_histogram
+        # Configure metric producers.
+        producers:
+          # Configure metric producer to be prometheus
+          - prometheus: {}
     # Configure a periodic metric reader.
     - periodic:
         # Configure exporter.
--- a/schema/meter_provider.json
+++ b/schema/meter_provider.json
@@ -33,6 +33,12 @@
                 },
                 "exporter": {
                     "$ref": "#/$defs/MetricExporter"
+                },
+                "producers": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/$defs/MetricProducer"
+                    }
                 }
             },
             "required": [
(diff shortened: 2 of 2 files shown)


schema/meter_provider.json
```diff
@@ -75,6 +87,27 @@
                 }
             }
         },
+        "MetricProducer": {
+            "type": "object",
+            "additionalProperties": true,
+            "minProperties": 1,
+            "maxProperties": 1,
+            "properties": {
+                "opencensus": {
```
dashpole, line 96: This was the only part I wasn't entirely sure about.  Since this is a pluggable interface, i've made them objects, with additional properties allowed.  I think that would allow adding configuration to these readers in the future.

schema/meter_provider.json
```diff
@@ -75,6 +87,27 @@
...
+                    "type": "object",
+                    "additionalProperties": false
+                },
+                "prometheus": {
```
dashpole, line 100: I wasn't sure whether to include the prometheus bridge here.  AFAIK this only exists in Go (maybe java too?).  Are properties added here required for implementations?  Or are they optional, and just a way to standardize the keys?

schema/meter_provider.json
```diff
@@ -33,6 +33,12 @@
                 },
                 "exporter": {
                     "$ref": "#/$defs/MetricExporter"
+                },
+                "producers": {
```
jack-berg, line 37: Let's add this as a property on `MetricReader` instead of PeriodicMetricReader and PullMetricReader.

schema/meter_provider.json
jack-berg: We only add schema surface area for components that are explicitly defined in the spec. The go implementation of this producer should be registered as a [ComponentProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#component-provider).

schema/meter_provider.json
jack-berg: As mentioned in another comment, we only add schema surface area for components which are explicitly definitely in the spec. Opencensus meets this critiera: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opencensus.md#metrics--stats

dashpole: OpenCensus is now the only one in this list.

schema/meter_provider.json
dashpole: I removed this component

schema/meter_provider.json
dashpole: Moved

schema/meter_provider.json
```diff
@@ -110,6 +133,12 @@
                 },
                 "pull": {
                     "$ref": "#/$defs/PullMetricReader"
+                },
+                "producers": {
```
jack-berg, line 137: Looks like we need to remove the `maxProperties` field. The goal for this was to constrain you to either specify pull or periodic, but not both. But with another property at the same level, we need to drop this constraint (or find a new way to enforce it).

examples/kitchen-sink.yaml
```diff
@@ -142,6 +142,10 @@ meter_provider:
               # Configure resource attributes to be excluded, in this example attribute service.attr1.
               excluded:
                 - "service.attr1"
+        # Configure metric producers.
+        producers:
```
jack-berg, line 146: I believe this is at the wrong level. (Applies to the other usage as well)
```suggestion
      producers:
```

schema/meter_provider.json
dashpole: removed

examples/kitchen-sink.yaml
dashpole: fixed here and below

examples/kitchen-sink.yaml
```diff
@@ -142,6 +142,10 @@ meter_provider:
               # Configure resource attributes to be excluded, in this example attribute service.attr1.
               excluded:
                 - "service.attr1"
+        # Configure metric producers.
```
codeboten, line 145:
```suggestion
      # Configure metric producers.
```

examples/kitchen-sink.yaml
```diff
@@ -203,6 +207,10 @@ meter_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION
             default_histogram_aggregation: base2_exponential_bucket_histogram
+        # Configure metric producers.
```
codeboten, line 210:
```suggestion
      # Configure metric producers.
```

schema/meter_provider.json
```diff
@@ -103,13 +126,18 @@
             "type": "object",
             "additionalProperties": false,
             "minProperties": 1,
-            "maxProperties": 1,
```
dashpole, line 106: Should I keep this, and just make it `"maxProperties": 2`

jack-berg: That seems fine for now.

jack-berg (approved):

schema/meter_provider.json
dashpole: As in: It is fine to change it to `"maxProperties": 2`, or the current state is fine?

jack-berg: Adding `maxProperties: 2` seems fine even if it doesn't perfectly provide the constraint we want.

dashpole: added

codeboten (approved):

schema/meter_provider.json
```diff
@@ -46,6 +46,12 @@
             "properties": {
                 "exporter": {
                     "$ref": "#/$defs/MetricExporter"
+                },
+                "producers": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/$defs/MetricProducer"
+                    }
                 }
```
bogdandrutu, line 55: Why is this needed here if already in the top level MetricReader? If it is required for pull (maybe to control at a granular level) why not for periodic as well?

dashpole: https://github.com/open-telemetry/opentelemetry-configuration/pull/93

jack-berg: Thanks for calling this out @bogdandrutu!

Issue #87, opened 2024-05-09
add support for instrumentation configuration?


brettmc:
Over in PHP SIG, we've started to implement file-based configuration, and it's going well.
We also have a requirement where some of our automatic instrumentation modules require configuration. So, following the file-configuration way of thinking, we're enabling this through a secondary .yaml file, along with another env var to point to the file. For example, `OTEL_PHP_INSTRUMENTATION_CONFIG_FILE=otel-instrumentation.yaml`:
```yaml
config:
  - example_instrumentation:
      span_name: ${EXAMPLE_INSTRUMENTATION_SPAN_NAME:-example span}
  - another_instrumentation:
      enabled: false
      foo_param: "some.value"
```

Could something like this be incorporated into the SDK configuration schema, for example under an `instrumentation` key?

jack-berg:
> Could something like this be incorporated into the SDK configuration schema, for example under an instrumentation key?

Yes I think so. There are a couple of things to think through:
- What configuration is language / framework agnostic enough to warrant inclusion in the schema?
- How should we accommodate language / framework specific configuration?
- How does instrumentation access the configuration? We talked about this in the 5/13/24 config working group and one idea was to add API surface area for configuration, such that instrumentation could access configuration from a global. We could expose something like the [component provider properties](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#create-plugin) type, and potentially some helper methods for instrumentation to resolve certain common configuration given a generic configuration type.

I think a good starting point is to include a handful of obvious generic instrumentation configuration properties in the schema, and work through some of these questions during that PR and in followups. One that @trask brought up was which headers should be captured for http semantic conventions. There are probably others.

Would you be willing to open a PR?

jack-berg:
#3560 is another example of instrumentation config.

jack-berg:
Related to https://github.com/open-telemetry/opentelemetry-specification/issues/3535.

jack-berg:
An old related issue where we tried to compile examples of instrumentation configuration: https://github.com/MrAlias/otel-schema/issues/8

brettmc:
> Would you be willing to open a PR?

Yes. I'll start to play around with it to see how I can achieve something basic first. I like the idea of having some generic/common properties - headers to capture is one that PHP also requires across a couple of different instrumentation packages.

Issue #89, opened 2024-05-16
Define stability guarantees for stable releases


jack-berg:
Fill in the details of the TODOs: https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#stability-definition

Should look to the [opentelemetry-proto stability guarantees](https://github.com/open-telemetry/opentelemetry-proto?tab=readme-ov-file#stability-definition) for guidance / prior art.

Need to find a way to have experimental portions of config even after the schema is stable overall, similar to how proto has experimental profile signal while logs, metrics, and traces are stable.

PR #91, opened 2024-05-20
adding instrumentation configuration

brettmc:
Initial draft of adding instrumentation configuration.
- added all of the instrumentation-specific examples I found in https://github.com/MrAlias/otel-schema/issues/8
- a `common` property for settings that are common to multiple instrumentations (ie, `otel.instrumentation.common.*`), and added schema for the ones I know about
- a series of language-specific properties, which are deliberately vague

Closes: #87

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -473,12 +473,21 @@ instrumentation:
         enabled: true
   # Language-specific module options
   java:
+    # Configure Java's Logback Appender instrumentation.
+    #
+    # See https://opentelemetry.io/docs/languages/java/automatic/spring-boot/#logback
     logback-appender:
+      experimental-log-attributes: true
       experimental:
+        capture-code-attributes: true
+        capture-marker-attribute: false
+        capture-key-value-pair-attributes: true
+        capture-logger-context-attributes: false
         capture-mdc-attributes:
           - attr_one
           - attr_two
   php:
+    # Configure PHP's example instrumentation.
     example_instrumentation:
       span_name: test123
       enabled: true
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -434,18 +434,23 @@ resource:
         - process.command_args
   # Configure the resource schema URL.
   schema_url: https://opentelemetry.io/schemas/1.16.0
+
 # Configure instrumentation.
 instrumentation:
   # Configure general SemConv options that may apply to multiple languages and instrumentations.
+  #
+  # Instrumenation may merge general config options with the language specific configuration at .instrumentation.<language>.
   general:
     # Configure instrumentations following the peer semantic conventions.
     # 
     # See peer semantic conventions: https://opentelemetry.io/docs/specs/semconv/attributes-registry/peer/
     peer:
       # Configure the service mapping for instrumentations following peer.service semantic conventions.
+      #
+      # Each entry is a key value pair where "peer" defines the IP address and "service" defines the corresponding logical name of the service.
       # 
       # See peer.service semantic conventions: https://opentelemetry.io/docs/specs/semconv/general/attributes/#general-remote-service-attributes
-      service-mapping:
+      service_mapping:
         - peer: 1.2.3.4
           service: FooService
         - peer: 2.3.4.5
@@ -456,46 +461,80 @@ instrumentation:
     http:
       # Configure instrumentations following the http client semantic conventions.
       client:
-        request:
-          capture-headers:
-            - X-Foo
-        response:
-          capture-headers:
-            - X-Foo
+        # Configure headers to capture for outbound http requests.
+        request_captured_headers:
+          - Content-Type
+          - Accept
+        # Configure headers to capture for outbound http responses.
+        response_captured_headers:
+          - Content-Type
+          - Content-Encoding
       # Configure instrumentations following the http server semantic conventions.
       server:
-        request:
-          capture-headers:
-            - X-Bar
-        response:
-          capture-headers:
-            - X-Bar
-    # Configure instrumentations following the db semantic conventions.
-    #
-    # See db semantic conventions: https://opentelemetry.io/docs/specs/semconv/database/
-    db:
-      # Configure database statement sanitization.
-      #
-      # See https://opentelemetry.io/docs/specs/semconv/database/database-spans/#common-attributes
-      statement-sanitizer:
-        enabled: true
-  # Language-specific module options
+        # Configure headers to capture for inbound http requests.
+        request_captured_headers:
+          - Content-Type
+          - Accept
+        # Configure headers to capture for outbound http responses.
+        response_captured_headers:
+          - Content-Type
+          - Content-Encoding
+  # Configure language-specific instrumentation libraries.
+  #
+  # Keys may refer to instrumentation libraries or collections of related configuration. Because there is no central schema defining the keys or their contents, instrumentation must carefully document their schema and avoid key collisions with other instrumentations.
+  #
+  # Configure C++ language-specific instrumentation libraries.
+  cpp:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure .NET language-specific instrumentation libraries.
+  dotnet:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Erlang language-specific instrumentation libraries.
+  erlang:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Go language-specific instrumentation libraries.
+  go:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Java language-specific instrumentation libraries.
   java:
-    # Configure Java's Logback Appender instrumentation.
-    #
-    # See https://opentelemetry.io/docs/languages/java/automatic/spring-boot/#logback
-    logback-appender:
-      experimental-log-attributes: true
-      experimental:
-        capture-code-attributes: true
-        capture-marker-attribute: false
-        capture-key-value-pair-attributes: true
-        capture-logger-context-attributes: false
-        capture-mdc-attributes:
-          - attr_one
-          - attr_two
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure JavaScript language-specific instrumentation libraries.
+  js:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure PHP language-specific instrumentation libraries.
   php:
-    # Configure PHP's example instrumentation.
-    example_instrumentation:
-      span_name: test123
-      enabled: true
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Python language-specific instrumentation libraries.
+  python:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Ruby language-specific instrumentation libraries.
+  ruby:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Rust language-specific instrumentation libraries.
+  rust:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Swift language-specific instrumentation libraries.
+  swift:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -485,56 +485,56 @@ instrumentation:
   #
   # Configure C++ language-specific instrumentation libraries.
   cpp:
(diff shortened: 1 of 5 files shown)


examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,37 @@ resource:
         - process.command_args
   # Configure the resource schema URL.
   schema_url: https://opentelemetry.io/schemas/1.16.0
+# Configure instrumentation
+instrumentation:
+  # Configure options common to multiple languages
+  common:
+    peer-service-mapping:
```
jack-berg, line 433: What do you think about organizing these by top level semconv namespace? So something like:
```
instrumentation:
  general:
    peer:
      service-mapping:
        - peer: foo
           service: bar
    http:
      client:
        captured-request-headers: 
          - X-Foo
        captured-response-headers: ... // omitted for brevity
      server:
        captured-request-headers: ... // omitted for brevity
        captured-response-headers: ... // omitted for brevity
  java: ..
  php: ..
```
Within a language, it makes sense to organize by instrumentation library like you have. This produces asymmetry, with the children of `instrumentation.general` being semconv namespaces, and `instrumentation.[language]` being instrumentation library names. One alternative could be to get rid of the language specific sections, and instead jump directly to configuration by instrumentation library names. With this, we could have a "semconv" configuration section, and instrumentation library configuration section:
```
instrumentation:
  semconv:
    peer:
      service-mapping:
        - peer: foo
           service: bar
    http:
      client:
        captured-request-headers: 
          - X-Foo
        captured-response-headers: ... // omitted for brevity
      server:
        captured-request-headers: ... // omitted for brevity
        captured-response-headers: ... // omitted for brevity
  libraries:
    logback-appender: ..
    example_instrumentation: ..
```
This could be problematic if the same instrumentation library key name is used in different languages but with different configuration schemas.

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,37 @@ resource:
...
+      - peer: foo
+        service: bar
+      - peer: bar
+        service: bat
+    http-capture-headers:
+      client:
+        request:
+          - X-Foo
+        response:
+          - X-Foo-*
```
jack-berg, line 443: We have guidance [here](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#schema-modeling-guidelines) for naming and semantics of properties which are expected to require pattern matching capabilities.

schema/instrumentation.json
```diff
@@ -0,0 +1,92 @@
...
+                "request": {
+                    "$ref": "#/$defs/ArrayOfHeaders"
+                },
+                "response": {
+                    "$ref": "#/$defs/ArrayOfHeaders"
+                }
+            },
+            "additionalProperties": false
+        },
+        "JavaModules": {},
```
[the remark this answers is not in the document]
jack-berg: I think we punt on that. Let's see how far we can take things with only including language agnostic instrumentation configuration options in the schema.

examples/kitchen-sink.yaml
brettmc: Organising general config to match semconv namespaces is a good idea, I've made that change.
My main concern with everything else being a free-for-all is conflict between keys across different languages, which we're unlikely to find out about until a user trips up on it.
A possible benefit of having a reserved space per language is that each language should be able to safely only process `general` plus their own language section for instrumentation config...?

examples/kitchen-sink.yaml
brettmc: Looking a bit more closely at [Java's documentation for capturing headers](https://opentelemetry.io/docs/languages/java/automatic/configuration/#capturing-http-request-and-response-headers), it doesn't look like wildcards are a feature, so I've just removed that for now.

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
         - process.command_args
   # Configure the resource schema URL.
   schema_url: https://opentelemetry.io/schemas/1.16.0
+# Configure instrumentation
+instrumentation:
+  # General SemConv options that may apply to multiple languages
+  general:
+    peer:
+      service-mapping:
```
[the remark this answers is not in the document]
brettmc: For peer service mapping, I can't find any usage or reference outside of Java, so it's OK by me to move it (under `java`) or remove it, until it's codified in the spec.

Capturing headers is mentioned in https://opentelemetry.io/docs/specs/semconv/http/http-spans/#common-attributes and I think the semconv-based way of representing the config avoids it being tied to any one implementation:

> Instrumentations SHOULD require an explicit configuration of which headers are to be captured

examples/kitchen-sink.yaml
jack-berg: > A possible benefit of having a reserved space per language is that each language should be able to safely only process general plus their own language section for instrumentation config...?

Yes that's true. Also, it makes explicit to the configuration author that they're configuring something language specific. Sure it results in an additional layer of nesting, but I think the positives outweigh this.

examples/kitchen-sink.yaml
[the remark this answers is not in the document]
jack-berg: > edit: https://opentelemetry.io/docs/specs/semconv/general/attributes/#general-remote-service-attributes hints that remote service mapping should be configurable by users:

Yup I found that same language. We can add a convention to state that we only add schema for general semconv options when they are expressed in the semantic conventions repo. This mirrors how we've been working thinking about schema coverage so far, where we only include configuration properties which are mentioned in the specification. Opened PR #92 to codify.

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
         - process.command_args
   # Configure the resource schema URL.
   schema_url: https://opentelemetry.io/schemas/1.16.0
+# Configure instrumentation
```
jack-berg, line 429:
```suggestion
# Configure instrumentation.
```

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
...
+instrumentation:
+  # General SemConv options that may apply to multiple languages
```
jack-berg, line 431:
```suggestion
  # Configure general SemConv options that may apply to multiple languages and instrumentations.
```

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
...
+  general:
+    peer:
```
jack-berg, line 433:
```suggestion
    # Configure instrumentations following the peer semantic conventions.
    # 
    # See peer semantic conventions: https://opentelemetry.io/docs/specs/semconv/attributes-registry/peer/
    peer:
```

examples/kitchen-sink.yaml
jack-berg, line 434:
```suggestion
      # Configure the service mapping for instrumentations following peer.service semantic conventions.
      # 
      # See peer.service semantic conventions: https://opentelemetry.io/docs/specs/semconv/general/attributes/#general-remote-service-attributes
      service-mapping:
```

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
...
+      service-mapping:
+        - peer: 1.2.3.4
+          service: FooService
+        - peer: 2.3.4.5
+          service: BarService
+    http:
```
jack-berg, line 439:
```suggestion
    # Configure instrumentations following the http semantic conventions.
    # 
    # See http semantic conventions: https://opentelemetry.io/docs/specs/semconv/http/
    http:
```

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
...
+      client:
```
jack-berg, line 440:
```suggestion
      # Configure instrumentations following the http client semantic conentions.
      client:
```

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,42 @@ resource:
...
+            - X-Bar
+        response:
+          capture-headers:
+            - X-Bar
+    db:
+      statement-sanitizer:
+        enabled: true
+  # Language-specific module options
+  java:
+    logback-appender:
```
jack-berg, line 459: Should we use scope names here? Seems to make sense instead of introducing an alternative key.

jack-berg (commented):
Left a few comments to add documentation to the updates to the `kitchen-sink.yaml`. Can you follow that comment format for all the new properties?

examples/kitchen-sink.yaml
brettmc: I think that's up to Java SIG. I've translated an example from https://github.com/MrAlias/otel-schema/issues/8#issuecomment-1291007692 but perhaps we just remove this example until it's decided how to represent Java's config?

examples/kitchen-sink.yaml
```diff
@@ -426,3 +426,68 @@ resource:
...
+          capture-headers:
+            - X-Bar
+    # Configure instrumentations following the db semantic conventions.
+    #
+    # See db semantic conventions: https://opentelemetry.io/docs/specs/semconv/database/
+    db:
+      # Configure database statement sanitization.
+      #
+      # See https://opentelemetry.io/docs/specs/semconv/database/database-spans/#common-attributes
+      statement-sanitizer:
```
brettmc, line 472: I think this is another instance of "hinted at by spec", in a footnote (5) for [db.statement](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#common-attributes):

> Should be collected by default only if there is sanitization that excludes sensitive information.

It doesn't mention whether users should be able to configure the behavior, though. In which case, maybe it gets the chop?

brettmc:
Just FYI, I'm going to be offline on vacation for a couple of weeks and won't be able to work on this.

jack-berg:
> Just FYI, I'm going to be offline on vacation for a couple of weeks and won't be able to work on this.

Not sure if you'll see this, but I'd like to pick this up and run with it if you don't mind me pushing some commits to your branch.

examples/kitchen-sink.yaml
trask: > Should we use scope names here? Seems to make sense instead of introducing an alternative key.

scope name is the "logger name" not the instrumentation library name

jack-berg: Yeah that kind of throws a wrench in that idea. Side note: it would be great if we could record information about the name / version of the log appender instrumentation library in scope attributes. Right now, that information is erased.

brettmc:
> Not sure if you'll see this, but I'd like to pick this up and run with it if you don't mind me pushing some commits to your branch.

No problem at all!

jack-berg:
> How do others feel about keeping the separator consistent using _? I worry that it would make usability more difficult if we support a mix of _ and -

👍 let's add it to [schema modeling rules](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#schema-modeling-rules)

jack-berg:
I pushed a commit to tweak a few things. Have a look folks! If we can get this merged, the next step is to start a conversation about figuring out how instrumentations can actually access this configuration via APIs.

jack-berg (approved):

examples/kitchen-sink.yaml
```diff
@@ -434,3 +434,107 @@ resource:
...
+    example:
+      property: "value"
+  # Configure PHP language-specific instrumentation libraries.
+  php:
+    # Configure the instrumentation corresponding to key "another_example_key".
+    example:
+      property: "value"
+  # Configure Python language-specific instrumentation libraries.
+  python:
+    # Configure the instrumentation corresponding to key "another_example_key".
```
brettmc, line 523:
```suggestion
    # Configure the instrumentation corresponding to key "example".
```

Repeated a couple of times.

jack-berg: Thanks

examples/kitchen-sink.yaml
```diff
@@ -434,3 +434,107 @@ resource:
...
+        - peer: 2.3.4.5
+          service: BarService
+    # Configure instrumentations following the http semantic conventions.
+    # 
+    # See http semantic conventions: https://opentelemetry.io/docs/specs/semconv/http/
+    http:
+      # Configure instrumentations following the http client semantic conventions.
+      client:
+        # Configure headers to capture for outbound http requests.
+        request_captured_headers:
```
trask, line 465: 😢 I like hyphens, but that's probably just a Java thing...

trask (approved):

schema/instrumentation.json
```diff
@@ -0,0 +1,134 @@
...
+        ".*": {
+            "$ref": "#/$defs/LanguageSpecificInstrumentation"
+        }
+    },
+    "$defs": {
+        "GeneralInstrumentation": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "peer": {
```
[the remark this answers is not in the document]
jack-berg: We need to solve the experimental vs. stable config properties in a general way. Tracking in #89.

examples/kitchen-sink.yaml
jack-berg: The casing of property keys should be consistent. While the decision of what case is somewhat arbitrary and subject to bikeshedding, but I think lower_snake_case is the best choice because its consistent with how the collector models YAML configuration today.

Opened #96 to codify this in naming rules.

jack-berg:
@open-telemetry/configuration-maintainers PTAL when you can. Thanks!

PR #94, opened 2024-05-28
add tool for config file validation and envvar replacement

tsloughter:


tsloughter:
Opening this even though I don't like the solution to the schema needing to not be a parent of the go file solution it uses of requiring use of a Makefile. But since there wasn't a meeting this week and I haven't gotten any responses on slack I figured I'd just open the existing solution and go from there with restructuring.

So I haven't bothered adding the github actions tests and publishing releases stuff.

tsloughter:
I marked this as ready for review because maybe that is why it wasn't getting any reviews...

I still don't like it though. I think the structure is wrong but wanted feedback, esp from Go users, on how to better structure it. But also in general if it would be ok that the code was just interspersed with the rest of the repo (as in, put `main.go` at the top level so it can read `schema/*` without it being copied to a subdirectory).

validator/main.go
```diff
@@ -0,0 +1,241 @@
...
+
+	return value
+}
+
+// Replace environment variables ${EXAMPLE} with their value and continue to
+// try replacing variables until there are no more, meaning ${EXAMPLE} could
+// contain another variable ${ANOTHER_VARIABLE}. But stop after 100 iterations
+// to prevent an infinite loop.
+// This does not use os.ExpandVars in order to support defaults like ${VAR:default}
+func expandString(s string) string {
```
marcalff, line 210: Did not test, but just from reading this comment, the behavior is violating the spec:

* https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution

> It MUST NOT be possible to inject environment variable by environment variables. For example, references to DO_NOT_REPLACE_ME environment variable below.

marcalff (commented):
Thanks for providing some tooling, looking to use it.

According to the spec, environment variable substitution should not be recursive.

validator/main.go
tsloughter: Oh, hm, I could have swore I included this because it was in the spec that you can do this. Thanks for calling it out, I'll remove the extra for-loop.

validator/LICENSE
jack-berg, line 1: I don't think a separate license file is needed.

validator/Makefile
jack-berg, line 1: Do you know if its idiomatic to have multiple make files in a single repo?

validator/README.md
```diff
@@ -0,0 +1,52 @@
...
+the schema:
+
+```
+$ make
+```
+
+Same is true for building the docker image:
+
+```
+$ make docker
```
jack-berg, line 19: Can you provide examples demonstrating usage of the docker image?

validator/go.mod
```diff
@@ -0,0 +1,16 @@
+module otel_config_validator
+
+go 1.22.0
+
+require (
+	github.com/santhosh-tekuri/jsonschema/v5 v5.3.1
```
jack-berg, line 6: @codeboten care to comment on the JSON schema tooling situation in the go ecosystem? I _believe_ the [go config implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/config/generated_config.go#L1C22-L1C57) uses github.com/atombender/go-jsonschema

validator/main_test.go
```diff
@@ -0,0 +1,51 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestExpandString(t *testing.T) {
```
jack-berg, line 9: Can you include tests for all the examples from the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution)?

validator/shelltests/help.test
```diff
@@ -0,0 +1,15 @@
+./otel_config_validator --help
```
jack-berg, line 1: What's going on with these `./validator/shelltests` files?

validator/main_test.go
jack-berg: Also, would be good to smoke test this by validating the files in `/examples`. Once we're confident in this tooling can replace the root make `validate-examples` task with this tool.

validator/LICENSE
tsloughter: Yup, was autogenerated and I didn't bother removing it because I'm still not happy with the layout since it requires copying the schema in a build step. But I can go ahead and remove this file.

validator/shelltests/help.test
tsloughter: They are just tests of the compiled cli. From the readme, `shelltest -c --diff --all shelltests/*.test`.

validator/go.mod
tsloughter: Ah interesting. I'm curious to know too. I'll look to jog my memory on why I went with `jsonschema`.

tsloughter: Oh, looks like its because github.com/santhosh-tekuri/jsonschema supports more validations. Maybe they are ones we don't need but I believe that is why I picked it.

validator/README.md
tsloughter: Will do.

validator/Makefile
tsloughter: It is as far as I know, but best to have the top level Makefile call this one -- which it doesn't right now.

The reason I didn't update the top level Makefile is I don't expect this one to remain here. I don't like requiring a makefile to build (`copy-schema` has to be run or the build fails or runs with an old copy of the schema) and think the go mod should be moved to the top level. I didn't do this because I wanted feedback first and wasn't sure if just requiring `make` was pretty idiomatic in Go anyway so no one would notice or care that the schema has to be copied.

tsloughter:
Update: I wasn't properly supporting types as the library was turning stuff like `disabled: ${OTEL_SDK_DISABLED:-false}` to `disabled: "false"`, but I think I've got it all resolved. Just gotta clean up my mess of Go code and will push up the fix tomorrow morning probably.

validator/go.mod
codeboten: I think i ended up with go-jsonschema because it supports code generation from the schema, either should be fine as long as they support the use-case they need

tsloughter:
I think issues with substitution have been resolved. Only thing is it now prefixes substituted values with the tag in yaml:

```
attribute_value_length_limit: !!int 0xdeadbeef
```

validator/main_test.go
tsloughter: Added a `make validator-validate-examples`.

validator/README.md
tsloughter: Added.

validator/main.go
tsloughter: Fixed.

validator/main_test.go
tsloughter: I haven't added a test for each example yet. But I've added some, like `validator/shelltests/hex_integer.test`.

I will add the rest but wanted to check if you were ok with using `shelltest` and `yq` to verify. It just feels simpler to test e2e that the output result is correct by just running it from the command line than dealing with Go tests. But that is likely in large part my inexperience with Go testing :).

tsloughter:
When this is accepted as the path forward I can add github action based publishing of cli binaries and docker image.

.github/workflows/validator-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+        go test .
+
+    - name: Run ShellTests
+      run: |
+        cd validator
+
+        sudo apt-get update
+        sudo apt-get install shelltestrunner
+
+        shelltest -c --diff --all shelltests/*.test
```
jack-berg, line 37: Is it feasible to include this in the makefile? Can `shelltest` be used on macos?

Makefile
```diff
@@ -2,6 +2,8 @@ SCHEMA_FILES := $(shell find . -path './schema/*.json' -exec basename {} \; | so
 EXAMPLE_FILES := $(shell find . -path './examples/*.yaml' -exec basename {} \; | sort)
 $(shell mkdir -p out)
 
+include validator/Makefile
+
 .PHONY: all
 all: install-tools compile-schema validate-examples
```
jack-berg, line 8: Should eventually extend all to include the appropriate validator tasks.

.github/workflows/validator-tests.yaml
jack-berg: It looks like `Dockerfile.shelltest` is aimed at running these tests inside a docker container. If that's what we expect users to run locally, we should probably run the same in the build for consistency.

validator/main_test.go
jack-berg: I'm ok using shelltest and yq as long as its easy to run locally outside of just linux machines.

validator/Makefile
jack-berg: We've got tooling in this project for js and now go. I think for sanity / consistency, it makes sense to bring all the commands together in a common framework like make.

tsloughter: @jack-berg ah, ok then.

validator/README.md
```diff
@@ -0,0 +1,74 @@
...
+Same is true for building the docker image:
+
+```
+$ make validator-docker
+```
+
+### Usage
+
+The command `otel_config_validator` takes one argument, the path to the yaml or
+json configuration file and optionally the path to a file to output the
```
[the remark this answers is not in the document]
tsloughter: Oh. I just assumed since it was defined as a jsonschema it meant json was supported.

I'm fine removing the json input support (except for the work I put into it, haha, but that is my fault for misunderstanding). 

However, what about json output? The reason I want json output is that it is more often found in standard libraries (that is the case now in Erlang/Elixir).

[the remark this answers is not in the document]
jack-berg: I don't have any problem with the program outputing a JSON representation as an intermediate step in the parsing process. Users should just be informed that as of now, the only file format they should use is YAML.

jack-berg:
> When this is accepted as the path forward I can add github action based publishing of cli binaries and docker image.

I think this is a reasonable direction. We can always iterate on the technologies used under the covers since they are abstracted away behind the docker interface. This provides a useful capability for end users and SDK implementations alike.

validator/README.md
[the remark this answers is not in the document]
tsloughter: Ok, json input support has been removed.

tsloughter:
@jack-berg ok, I've added binary and docker publish support to the github actions.

Docker images are published to ghcr. Binary releases are built on tags with a tool called `cargo-dist` (originally built for building binaries with `cargo` but now supports generic commands to compile the binary, I've been using this on a couple other projects recently).

Open questions:

* Do we want a homebrew recipe? `cargo-dist` will generate it and publish to a tap if we get `open-telemetry/homebrew-tap` repo created. I figured unnecessary to start, but it can easily be added if deemed useful and they agree to create the repo.
* Should I get rid of the install script and the user just have to download the right binary? My main issue with the install script is you need a path to install the binary and there is no `XDG_` support builtin. What I used is `"[$OTEL_CONFIG_VALIDATOR_HOME/bin", "~/.otel_config_validator/bin"]`, which means if user defines `$OTEL_CONFIG_VALIDATOR_HOME` it will install there otherwise `"~/.otel_config_validator/bin"` -- but I hate polluting `~/` with dotfiles.
* I haven't tested Powershell yet. I may just remove for now and we can add it back later if needed?
* Docker images are published to ghcr. Does it need to go to docker hub or like a CNCF quay or something?
* Docker images are published with docker tags of the full version `major.minor.patch`, plus  `major.minor` and `major`. Should we only publish for the full version?
* Images are published on merge to `main` with just the sha git ref as the docker tag. This can make it simpler for people to test between releases but does greatly increase the number of published images and may be unnecessary. Should I remove it?

jack-berg:
@tsloughter can we go through these in person at the next config SIG meeting on 8/5? Not sure if you can make it, but seems like it'd be easier than working through these questions async.

tsloughter:
@jack-berg yes, and agreed. I can do my best to make the meeting on Monday. Likely can only do half the meeting, but won't know which half until closer to start time...

codeboten (commented):
Would it be useful to use the same tools as the collector uses for shipping binaries of the collector? goreleaser has worked pretty well and does simplify the github action configuration needed

You can see an example of the configuration https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.github/workflows/base-ci-goreleaser.yaml

I would also be happy to help move to that tool if its deemed useful

tsloughter:
@jack-berg I finally updated to accept a schema as an option to the cli. Also broke the version from versioning with the schema and added a prefix `validator-` required for tags that will trigger publishing of the releases and its assets.

Something I'd like to add, but can be added after a merge, is a command to print the schema version that is embedded in the tool. But it doesn't seem the version is anywhere in the schema itself? A version of the schema is based solely on the git tag it is associated with? That makes it tough.. I was hoping it was just a field I could read from `schema/opentelemetry_configuration.json`. I suppose I could also hardcode it into `main.go` and require it be updated each time a new schema is released.

tsloughter:
@codeboten I'm fine with whatever. I used cargo-dist because I had previous experience with it from another project and really liked it, then used it again on a Go project so knew it worked for those.

I take it there is a Pro plan that could be used, since most features are behind the Pro plan (ones being used here with cargo-dist)?

At least looks like it? https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.github/workflows/base-ci-goreleaser.yaml#L69

codeboten:
> I take it there is a Pro plan that could be used

Right, we use the pro version for the collector releases. I don't have a strong preference other than that i'm more familiar with goreleaser 😅

.github/workflows/validator-release.yml
```diff
@@ -0,0 +1,281 @@
+# Copyright 2022-2024, axodotdev
```
jack-berg, line 1: This file is pretty dense and difficult to review with certainty. Was it based off some existing release workflow?

.github/workflows/validator-docker.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Build and push
+      id: push
+      uses: docker/build-push-action@v6
+      with:
+        context: ./validator/
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
+        file: ./validator/Dockerfile
+        push: ${{ github.event_name != 'pull_request' }}
```
jack-berg, line 60: Why not just on run this on pushes to main?

jack-berg (approved):
I have a few small lingering questions but this seems like a great start.

validator/README.md
```diff
@@ -0,0 +1,92 @@
...
+### Testing
+
+Run the Go unit tests:
+
+```
+$ go test .
+```
+
+Running the tests of the compiled CLI requires
+[shelltest](https://github.com/simonmichael/shelltestrunner),
```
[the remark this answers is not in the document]
tsloughter: You can use `brew` or the docker image `Docker.shelltest`.

.github/workflows/validator-release.yml
tsloughter: It is generated by `cargo-dist`.

validator/README.md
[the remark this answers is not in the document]
jack-berg: Ah they don't advertise brew from the github repo

validator/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+RUN --mount=type=cache,target=/root/.cache/go-build,id=cache-go-build --mount=type=cache,target=/root/.cache/go-mod,id=cache-go-mod \
+    CGO_ENABLED=0 GOOS=linux go build -o /otel_config_validator
+
+FROM gcr.io/distroless/base-debian12
```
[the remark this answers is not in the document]
tsloughter: A few reasons for that chose, one of which I know learn is false. The first one being that since I was inexperienced with Go I googled around to see what was the common base for a runner and it appeared to be this. That may be true.

Second, I thought it was smaller than alpine -- turns out its not. `static-debian12` i smaller (2MB) but `base-debian12` is actually 20MB. I can look into if static would be a good choice.

Last, I have just steered away from alpine for years now. It is my initial instinct to choose debian after seeing too many issues. That isn't to say alpine wouldn't be fine in this case, just the reason I personally didn't default to alpine. 

I'm happy with whatever, but would defer to what is the defacto standard (if any) for Go projects.

validator/README.md
```diff
@@ -0,0 +1,92 @@
+## OpenTelemetry SDK Configuration Validator
+
+This application will replace environment variables in values of valid yaml
+files, following the rules of [file configuration environment variable
+substitution](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution),
+before validating that result against a version of the [OpenTelemetry SDK
+Configuration
+schema](https://github.com/open-telemetry/opentelemetry-configuration/).
```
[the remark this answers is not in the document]
tsloughter: Updated.

validator/README.md
```diff
@@ -0,0 +1,92 @@
...
+```
+$ docker run -v $(pwd):/opt/otel_config_validator otel_config_validator:current -o out.yaml examples/kitchen-sink.yaml
+```
+
+With the above docker command the output file, `out.yaml`, will be owned by
+`root:root` but be readable by any user.
+
+Environment variable substitution is supported with the syntax `${VARIABLE}` or
+`${env:VARIABLE}`. Default values are supported in the form
+`${VARIABLE:-default}`.
```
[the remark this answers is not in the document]
tsloughter: Yup, linked to it in the opening, but agree its good to have here too. Updated.

.github/workflows/validator-docker.yaml
tsloughter: I think the thinking was because it has to run both on `push` and `release` but the job runs for `push`, `pull_request` and `release`. Not that it couldn't be done as an `if push or release`.

validator/Makefile
```diff
@@ -0,0 +1,24 @@
...
+DOCKER_IMAGE_TAG := ${CURRENT_GIT_REF}
+DOCKER_BUILD_ARGS := -f ${ROOT_DIR}/Dockerfile -t otel_config_validator:${DOCKER_IMAGE_TAG} -t otel_config_validator:current
+EXAMPLE_FILES := $(shell find ${ROOT_DIR}/../examples -name "*.yaml" -exec basename {} \; | sort)
+$(shell mkdir -p out)
+
+validator-copy-schema:
+	cp -R ${SCHEMA_DIR} ${ROOT_DIR}/
+
+validator: validator-copy-schema
+	go build -C ${ROOT_DIR} ${ROOT_DIR}
```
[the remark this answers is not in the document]
tsloughter: Updated to note 1.20+ is required.

validator/Dockerfile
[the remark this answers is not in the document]
jack-berg: > I'm happy with whatever, but would defer to what is the defacto standard (if any) for Go projects.

Agreed - I prefer following some standard vs inventing / justifying a new one.

The collector releases use alpine: https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol/Dockerfile#L1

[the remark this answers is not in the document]
tsloughter: Oh, interesting. I guess we should follow their lead. Unless @MrAlias or @codeboten suggest otherwise to be the Go defactor standard?

tsloughter:
@jack-berg I updated the Dockerfile to build in the binary and the shelltests to be able to use a bin from anywhere (not just `./otel_config_validator`. As well as the github action to use a new makefile target that builds/runs the docker image to execute the shelltests. Plus, updated the readme to show using the makefile to run the shelltests (so it uses docker as well, no need to install anything else).

validator/Dockerfile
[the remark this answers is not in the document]
codeboten: The collector uses alpine for building but then switches to `scratch` which is pretty standard for go tools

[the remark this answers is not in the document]
jack-berg: @tsloughter do you want to switch to `scratch` and then merge this?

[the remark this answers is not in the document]
tsloughter: @jack-berg yup, will do

PR #97, opened 2024-06-14
add exporter retry configuration

brettmc:
https://opentelemetry.io/docs/specs/otlp/#otlphttp-throttling and https://opentelemetry.io/docs/specs/otlp/#otlpgrpc-throttling describe how a client SHOULD implement an exponential backoff strategy (with jitter) in case of retryable export failures. The inputs to this strategy are usually the initial delay, and max attempts.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -91,6 +91,12 @@ logger_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
             insecure: false
+            # Configure retry policy
+            retry:
+              # Configure maximum number of retries before failing.
+              max_attempts: 3
+              # Configure initial delay (in milliseconds) to be used for exponential backoff with jitter.
+              initial_delay: 100
     # Configure a simple span processor.
     - simple:
         # Configure exporter.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+ * Add exporter `retry` configuration.
+
 ## [v0.2.0] - 2024-05-08
 
 * Document time value units in kitchen-sink example. [#51](https://github.com/open-telemetry/opentelemetry-configuration/pull/51)
--- a/schema/common.json
+++ b/schema/common.json
@@ -63,6 +63,9 @@
                 },
                 "insecure": {
                     "type": "boolean"
+                },
+                "retry": {
+                    "$ref": "#/$defs/Retry"
                 }
             },
             "required": [
@@ -74,6 +77,19 @@
         "Console": {
             "type": "object",
             "additionalProperties": false
+        },
+        "Retry": {
+            "type": "object",
+            "properties": {
+                "max_attempts": {
+                    "type": "integer",
+                    "minimum": 0
+                },
+                "initial_delay": {
+                    "type": "integer",
+                    "minimum": 1
+                }
+            }
         }
     }
 }


weslenteche (approved):

schema/meter_provider.json
```diff
@@ -166,6 +166,12 @@
                     "type": "integer",
                     "minimum": 0
                 },
+                "insecure": {
```
brettmc, line 169: moved `insecure` up to be consistent with common.json/Otlp

codeboten (commented):
Thanks @brettmc for the PR! I wonder how consistent the implementations of this retry are. In Go the options are presented slightly differently:

```go

	// Enabled indicates whether to not retry sending batches in case of
	// export failure.
	Enabled bool
	// InitialInterval the time to wait after the first failure before
	// retrying.
	InitialInterval time.Duration
	// MaxInterval is the upper bound on backoff interval. Once this value is
	// reached the delay between consecutive retries will always be
	// `MaxInterval`.
	MaxInterval time.Duration
	// MaxElapsedTime is the maximum amount of time (including retries) spent
	// trying to send a request/batch.  Once this value is reached, the data
	// is discarded.
	MaxElapsedTime time.Duration
```

And in Python, the only option is a max interval:

```python
def _create_exp_backoff_generator(max_value: int = 0) -> Iterator[int]:
    """
    Generates an infinite sequence of exponential backoff values. The sequence starts
    from 1 (2^0) and doubles each time (2^1, 2^2, 2^3, ...). If a max_value is specified
    and non-zero, the generated values will not exceed this maximum, capping at max_value
    instead of growing indefinitely.

    Parameters:
    - max_value (int, optional): The maximum value to yield. If 0 or not provided, the
      sequence grows without bound.
```

brettmc:
> I wonder how consistent the implementations of this retry are

It looks like not at all consistent, and I just found https://github.com/open-telemetry/opentelemetry-specification/issues/3639 which describes it quite well:

> The implementation of the exponential backoff strategy is not specified

It seems like the root of the issue then is that we cannot configure retry consistently across SIGs if we haven't agreed on the inputs. I can think of a couple of possible next steps:

- collate all inputs in use by all of our implementations of retry policy, and allow all of those fields to be part of retry config (with SDKs choosing the fields they are interested in). It would work, but the same file-based configuration will lead to different behaviour across SIGs, which is not really in the spirit of this repo.
- go back to the spec and try to come up with a specification for how retry should be configured, which everybody can agree on

svrnm:
> go back to the spec and try to come up with a specification for how retry should be configured, which everybody can agree on

This sounds like the best path forward, with the other option being the fallback. @brettmc would you mind raising the spec issue, with 

* a reference to this issue
* a reference to how PHP and Go does it

FYI, this is something related to https://github.com/open-telemetry/opentelemetry-specification/issues/4083

jack-berg:
To add to @codeboten analysis of existing implementations, here is [java's set of options](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/RetryPolicy.java)

- maxAttempts
- initialBackoff
- maxBackoff
- backoffMultiplier

These options mirror [gRPC's built in retry mechanism](https://grpc.io/docs/guides/retry/#retry-configuration), since gRPC was very influential in the development of OTLP and it seemed reasonable that OTLP clients using the gRPC version of the protocol would utilize the gRPC clients' built in mechanism.

One concrete thing we can / should do while we sort out how these options are configurable across languages is simply give the ability to enable / disable retry. Regardless of the options, every langauge should support _some_ retry mechanism since the [spec requires it in clear terms](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry):

> Transient errors MUST be handled with a retry strategy. This retry strategy MUST implement an exponential back-off with jitter to avoid overwhelming the destination until the network is restored or the destination has recovered.

We could do this by providing a top level `retry_enabled` property for the OTLP exporter type:

```
otlp:
  endpoint: http://localhost:4318
  retry_enabled: true
  # add in the future
  # retry_options:
```

Later when we add specific options we could add a separate `retry_options` type.

Or we could define a new `retry` type, with only a `enabled` property to start, with the intent to expand later when we decide on the options:

```
otlp:
  endpoint: http://localhost:4318
  retry:
    enabled: true
    # add additional options in the future
```

brettmc:
> Or we could define a new retry type, with only a enabled property to start, with the intent to expand later when we decide on the options

I like this one, combining all retry-related options into one type which we can expand on.

jack-berg:
> I like this one, combining all retry-related options into one type which we can expand on.

👍 This sounds good to me. Do you have interest in updating this PR to that effect? Or did you want to wait until the specific options are worked out at the spec lever (which could be some time).

brettmc:
> Do you have interest in updating this PR to that effect?

Updated to only implement retry.enabled (boolean, default true). The rest of the configuration will need to wait until there is agreement on what inputs are required.

examples/kitchen-sink.yaml
```diff
@@ -91,6 +91,10 @@ logger_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
             insecure: false
+            # Configure retry policy
```
jack-berg, line 94: #nit for style symmetry. 
```suggestion
            # Configure retry policy.
```

examples/kitchen-sink.yaml
```diff
@@ -199,6 +203,10 @@ meter_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_METRICS_INSECURE
             insecure: false
+            # Configure retry policy
+            retry:
+              # Configure retry enabled
```
jack-berg, line 208: Not sure this is the PR to solve it, but we might need to come up with a strategy for making it clear what the default values are. Our guidance in the [spec is](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#create):

> If a field is null or unset and a default value is defined, Create MUST ensure the SDK component is configured with the default value. 

In every case so far, its always been clear what the default is because we have a policy [that says we only include properties](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#what-properties-are-part-of-schema) in the schema which are part of the spec. This means SDKs can always check the spec to see what the default is. 

In the case of retry enabled, the [spec says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry): 

> Transient errors MUST be handled with a retry strategy. 

This suggests that `*.retry.enabled=true` by default, but I'm not sure that's the case in most SDK implementations.

examples/kitchen-sink.yaml
```diff
@@ -91,6 +91,10 @@ logger_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
             insecure: false
+            # Configure retry policy
+            retry:
+              # Configure retry enabled
+              enabled: true
```
jack-berg, line 97: If we buy my argument that `enabled: true` by default, then we should rename it to `disabled` to align with [spec naming conventions for booleans](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#boolean-value):

> All Boolean environment variables SHOULD be named and defined such that false is the expected safe default behavior.

examples/kitchen-sink.yaml
jack-berg: Opened https://github.com/open-telemetry/opentelemetry-specification/pull/4148 to explicitly define a corresponding option for this in the spec along with a default.

examples/kitchen-sink.yaml
```diff
@@ -91,6 +91,10 @@ logger_provider:
             #
             # Environment variable: OTEL_EXPORTER_OTLP_INSECURE, OTEL_EXPORTER_OTLP_LOGS_INSECURE
             insecure: false
+            # Configure retry policy.
+            retry:
+              # Configure retry disabled.
+              disabled: false
```
jack-berg, line 97: What do you think of @carlosalberto's [suggestion](https://github.com/open-telemetry/opentelemetry-specification/issues/4138#issuecomment-2239590144) to have a `strategy` field, defaulting to `exponential_retry` and supporting `none`.

brettmc: I really like it, and I think it could resolve my issues. If we allow for future retry strategies, including custom/3rd party strategies, then SIGs have the freedom to provide their own configuration to suit their specific implementation.
If the official exponential retry strategy is locked down in spec, SIGs can migrate across to it at their leisure (w.r.t using it with file-based configuration).

How about treating retry policy similarly to how we treat a `Sampler` - it's configuration is optional and defaults to `exponential_retry`, at most one can be provided, and custom is allowed? Something like:

```yaml
retry:
  disabled: false
  <policy-name>:
    property: value
```

`disabled` could actually be replaced by a ` noop`/`none`/`noretry` policy, which might make the interface between an exporter and the retry policy a little cleaner.

examples/anchors.yaml
```diff
@@ -12,6 +12,8 @@ exporters:
       api-key: !!str 1234
     compression: gzip
     timeout: 10000
+    retry:
+      disabled: false
```
zeitlinger, line 16: `enabled` seems to be easier - avoids double negation

jack-berg:
@brettmc it doesn't seem like there's consensus to add the corresponding option in the spec (spec [PR #4148](https://github.com/open-telemetry/opentelemetry-specification/pull/4148)). If you know of use cases where PHP users want to disable retry, please comment on that PR. 

Else, it seems like we are blocked on this until specification issues are resolved.

examples/anchors.yaml
tsloughter: The spec defines that options like this have to be `disabled` and not `enabled`. But I can't find it right now... so I may be wrong.

tsloughter: Oh, just saw Jack already linked to it earlier, https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#boolean-value

PR #104, opened 2024-07-19
Project tooling to generate example property descriptions

jack-berg:
## Problem

This project maintains [several examples](https://github.com/open-telemetry/opentelemetry-configuration/tree/main/examples). These serve multiple purposes
- Validate correctness of schema changes
- Serve as templates for users
- Document the semantics about configuration properties for both users who set them, and maintainers who implement the interpretation of these properties. 

The last point is poorly served. The comments on these examples contain important information, like the priority between `.attribute_limits` and `.tracer_provider.limits`, the set of recognized values for samplers, the set of recognized values for view stream aggregations, and more. Trying to ensure that comments are maintained consistently across examples is time consuming and error prone.

## Solution

This PR introduces new project tooling to generate the comments consistently across example files.

How it works:

* The `./schema/type_descriptions.yaml` file defines descriptions for each of the properties of each type defines in the `./schema` data model.
* The `./scripts/generate-comments.js` is a script which for a given input configuration file will:
  * Parse the YAML.
  * Walk through each key / value pair, and for each:
    * Compute the JSON dot notation location of the current key / value pair.
    * Find the first matching rule in `type_description.yaml`. Iterate through the rules and evaluate the key / value pair dot notation location against each of the rule's `path_patterns`.
    * Inject / overwrite comments for its properties according to `type_descriptions.yaml`.
  * Write the resulting content to specified output file or to the console.

The `make generate-descriptions` command runs this process against each file in `./examples` and overwrites each file in the process. 

## Alternatives considered

Can't we just use the [JSON schema description field](https://json-schema.org/learn/getting-started-step-by-step#create-a-schema-definition) to encode this information? 

Not easily, and with some drawbacks. In this PR I use a yaml parsing library to visit each key / value pair, evaluate its comment, and update the yaml parsing libraries in-memory representation of the comment for that pair. If the description information was encoded in the JSON schema, I'd need to combine capabilities from two libraries: one js library for parsing / validating yaml according to the JSON schema, and another js library for visiting each node and updating representation. The JSON schema parsing library doesn't provide tools for mutating the state of the YAML comments. I don't see an obvious way to combine the necessary functionality from each of these libraries.

Additionally, JSON schema is written in JSON, which has poor support for the types of multi-line comments which are going to be common.

This approach has the drawback that the knowledge about a particular type is distributed across two places: the `.json` JSON schema file and the `type_descriptions.yaml` file. But in exchange for this, we get a simple script for tooling, and its ready to use today 🙂 .


jack-berg:
Per our convo in the 7/22/24 config SIG, I've pushed a commit to run `make generate-descriptions` and check for diff in the build.

scripts/generate-comments.js
```diff
@@ -0,0 +1,171 @@
...
+        const matchingRule = typeDescriptionsYaml.find((rule) => {
+            const matchingPathPattern = rule['path_patterns'].find((pathPattern) => {
+                const regex = new RegExp(toRegex(pathPattern));
+                return regex.test(parentPath);
+            });
+            return matchingPathPattern !== undefined;
+        });
+        // Exit early if no matching rule
+        if (matchingRule === undefined) {
+            maybeDebugLog("no matching rule")
```
zeitlinger, line 93: is this an error - do we want comments for all entries?

scripts/generate-comments.js
```diff
@@ -0,0 +1,171 @@
...
+            return;
+        }
+        maybeDebugLog("matched rule: " + matchingRule.type);
+        // Check if there is a description for the current propertyKey in the matching rule
+        // Exit early if none registered
+        const description = matchingRule['property_descriptions'][propertyKey];
+        if (description === undefined) {
+            maybeDebugLog("no matching property")
```
zeitlinger, line 101: is this an error - do we want comments for all entries?

scripts/generate-comments.js
```diff
@@ -0,0 +1,171 @@
...
+    let parts = pattern.split("*");
+    if (parts.length === 0) parts = [pattern];
+    const escaped = parts
+        .map(chunk => chunk.replace(/[-[\]{}()*+?.,\\^$|]/g, "\\$&"))
+        .join("(.*)");
+    return "^" + escaped + "$";
+}
+
+// Log the message to the console if the script was run with `--debug` argument
+function maybeDebugLog(message) {
```
zeitlinger, line 152:
```suggestion
function debug(message) {
```

scripts/generate-comments.js
jack-berg: Yeah let's go with that. Easier to start with stronger assertions and relax later than it is to go the other way around.

jack-berg: Turns out this is too restrictive since there are cases where a file's contents do not conform to any schema. For example, [anchors.yaml](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/anchors.yaml#L4-L13) demonstrates the usage of YAML anchors, which don't match to any type in the schema until they are resolved.

scripts/generate-comments.js
jack-berg: Same [comment](https://github.com/open-telemetry/opentelemetry-configuration/pull/104#discussion_r1721910849) applies here.

jack-berg:
I've resolved merge conflicts. @open-telemetry/configuration-approvers please take another look. Im happy to iterate as needed, but we do need tooling to enforce consistent property descriptions.

codeboten (approved):
Needs one more conflict resolution, otherwise looks good

PR #106, opened 2024-07-22
Define modeling guidance for mismatches with standard env vars, add resource.attribute_list

jack-berg:
Resolves https://github.com/open-telemetry/opentelemetry-specification/issues/3966.

- Define modeling guidance that properties should use the most appropriate data types for the problem domain
- This may result in the lack of ability to support standard env vars where type mismatches occur
- When there is a type mismatch, a demonstrated need for platforms to contribute to config, and no reasonable alternative, we provide alternative versions of the property which match the standard env var
- For example, `OTEL_RESOURCE_ATTRIBUTES` is string whose value is a comma separated list of key value pairs. This is not a natural way to represent key value pairs in YAML, yet for reasons discussed in [PR#3966](https://github.com/open-telemetry/opentelemetry-specification/issues/3966) is important for platforms. Therefore, we provide an alternative `.resource.attributes_list` property which matches the format of `OTEL_RESOURCE_ATTRIBUTES`. Implementations merge the contents of `.resource.attributes` and `.resource.attributes_list`, with `.resource.attributes` taking priority.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This repository contains the JSON schema that defines the OpenTelemetry configur
 
 The [examples](./examples) repository contains a variety of sample configuration files to help get started and illustrate useful patterns. The following are noteworthy:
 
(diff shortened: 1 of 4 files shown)


zeitlinger (approved):
lgtm

jsuereth:
FYI - I'd like to go a different direction with resource variables BUT I think this makes good progress on status quo.

See https://github.com/open-telemetry/oteps/pull/264/files - environment variable detector

codeboten (approved):

brettmc (approved):

PR #107, opened 2024-07-24
Configure exporter default aggregation and temporality preference

jack-berg:
With https://github.com/open-telemetry/opentelemetry-specification/pull/4142 merged, the spec now has explicit configuration options for each metric exporter's temporality preference and default temporality.

This PR extends the metric exporters with new properties:

- console exporter has new `temporality_preference` and `default_aggregation_by_instrument_kind` properties
- prometheus exporter has new `default_aggregation_by_instrument_kind` property
- otlp exporter's `default_histogram_aggregation` property migrates to new more expressive `default_aggregation_by_instrument_kind` property

This supersedes #99.

### DIFF
--- a/examples/anchors.yaml
+++ b/examples/anchors.yaml
@@ -32,7 +32,9 @@ meter_provider:
             <<: *otlp-exporter
             endpoint: http://localhost:4318/v1/metrics
             temporality_preference: delta
-            default_histogram_aggregation: base2_exponential_bucket_histogram
+            default_aggregation_by_instrument_kind:
+              histogram:
+                base2_exponential_bucket_histogram:
 
 tracer_provider:
   processors:
--- a/schema/meter_provider.json
+++ b/schema/meter_provider.json
@@ -63,7 +63,7 @@
                     "$ref": "#/$defs/OtlpMetric"
                 },
                 "console": {
-                    "$ref": "common.json#/$defs/Console"
+                    "$ref": "#/$defs/ConsoleMetric"
                 },
                 "prometheus": {
                     "$ref": "#/$defs/Prometheus"
@@ -113,6 +113,9 @@
                 },
                 "with_resource_constant_labels": {
                     "$ref": "common.json#/$defs/IncludeExclude"
+                },
+                "default_aggregation_by_instrument_kind": {
+                    "$ref": "#/$defs/DefaultAggregationByInstrumentKind"
                 }
             }
         },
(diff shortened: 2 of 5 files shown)


schema/meter_provider.json
```diff
@@ -186,6 +185,116 @@
...
+                },
+                "observable_up_down_counter": {
+                    "$ref": "#/$defs/AggregationTemporality"
+                },
+                "up_down_counter": {
+                    "$ref": "#/$defs/AggregationTemporality"
+                }
+            }
+        },
+        "AggregationTemporalityPreference": {
```
jack-berg, line 188: The spec says that aggregation temporality should be configurable as a function of instrument kind, but most users will want to use these [shorthand options](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md#additional-environment-variable-configuration). 

This is much simpler / more compact than writing the preference for each instrument kind. In the future, we can add a separate property to give full control over temporality by instrument kind if needed.

schema/meter_provider.json
```diff
@@ -186,6 +185,116 @@
...
+                },
+                "observable_up_down_counter": {
+                    "$ref": "#/$defs/Aggregation"
+                },
+                "up_down_counter": {
+                    "$ref": "#/$defs/Aggregation"
+                }
+            }
+        },
+        "Aggregation": {
```
jack-berg, line 220: Lifting this aggregation type def up from the view stream definition. Its now reused both for view stream config and in exporter default aggregation config.

examples/anchors.yaml
```diff
@@ -32,7 +32,9 @@ meter_provider:
             <<: *otlp-exporter
             endpoint: http://localhost:4318/v1/metrics
             temporality_preference: delta
-            default_histogram_aggregation: base2_exponential_bucket_histogram
+            default_aggregation_by_instrument_kind:
```
zeitlinger, line 35: what if I want to
- have an aggregation preference for all kinds?
- have settings that apply for all exporters?

I see that there is a yaml anchor - but I'm not sure this should be the only option to share values

tsloughter: Using anchors makes using the config easier. No need to merge properties from some global exporter config with per-exporter config. And simplifies the schema.

zeitlinger: it's a lesser known and complicated feature

jack-berg:
Closing this. Can revisit if there's more demand for configuration outside of the defaults.

PR #111, opened 2024-08-19
fix: update attribute_keys to include type

### LINKED ISSUE #98: Should attribute_keys in stream configuration support include/exclude syntax?
The specification states:

> attribute_keys: This is, at a minimum, an allow-list of attribute keys for measurements captured in the metric stream. The allow-list contains attribute keys that identify the attributes that MUST be kept, and all other attributes MUST be ignored.

But looking at the go implementation, an attribute filter can be specified either to include or exclude based on the attribute keys:

https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/metric/instrument.go#L146

I checked the python implementations, and it looks like that implementation only supports including attributes based on the keys.

codeboten:
This updates the stream's `attribute_keys` configuration to an include/exclude style object that currently only supports `included`. This is in hope to support `excluded` in the future once the specification allows it.

Fixes #98

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Adding initial instrumentation configuration schema
 * Split MetricExporter into PullMetricExporter and PushMetricExporter and ensure only PushMetricExporters can be associated with PeriodicMetricReader [#110](https://github.com/open-telemetry/opentelemetry-configuration/pull/110)
-* Update `attribute_keys` from array to to Include type
+* Update `attribute_keys` from array to to Include type [#111](https://github.com/open-telemetry/opentelemetry-configuration/pull/111)
 
 ## [v0.2.0] - 2024-05-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Adding initial instrumentation configuration schema
 * Split MetricExporter into PullMetricExporter and PushMetricExporter and ensure only PushMetricExporters can be associated with PeriodicMetricReader [#110](https://github.com/open-telemetry/opentelemetry-configuration/pull/110)
+* Update `attribute_keys` from array to to Include type
 
 ## [v0.2.0] - 2024-05-08
 
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -249,8 +249,10 @@ meter_provider:
             record_min_max: true
         # Configure attribute keys retained in the resulting stream(s).
         attribute_keys:
-          - key1
-          - key2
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Adding initial instrumentation configuration schema
 * Split MetricExporter into PullMetricExporter and PushMetricExporter and ensure only PushMetricExporters can be associated with PeriodicMetricReader [#110](https://github.com/open-telemetry/opentelemetry-configuration/pull/110)
+* Update `attribute_keys` from array to to Include type
```
codeboten, line 7:
```suggestion
* Update `attribute_keys` from array to to Include type [#111](https://github.com/open-telemetry/opentelemetry-configuration/pull/111)
```

examples/kitchen-sink.yaml
```diff
@@ -249,8 +249,10 @@ meter_provider:
             record_min_max: true
         # Configure attribute keys retained in the resulting stream(s).
         attribute_keys:
-          - key1
-          - key2
+          # Configure list of attribute key patterns attribute keys retained in the resulting stream(s).
```
MrAlias, line 252:
```suggestion
          # Configure list of attribute key patterns that are retained in the resulting stream(s).
          # All other attributes not matching will be dropped.
```

MrAlias (approved):

examples/kitchen-sink.yaml
```diff
@@ -249,8 +249,11 @@ meter_provider:
             record_min_max: true
         # Configure attribute keys retained in the resulting stream(s).
         attribute_keys:
-          - key1
-          - key2
+          # Configure list of attribute key patterns that are retained in the resulting stream(s).
+          # All other attributes not matching will be dropped.
+          included:
```
jack-berg, line 254: The expansion from absolute values to patterns extends beyond the language of the spec. My understanding is that this adjustment is to help improve ergonomics for when the spec also includes the ability to exclude values. Can we start with absolute keys and evolve to allow patterns if the spec adds support?

examples/kitchen-sink.yaml
```diff
@@ -249,8 +249,11 @@ meter_provider:
             record_min_max: true
         # Configure attribute keys retained in the resulting stream(s).
         attribute_keys:
-          - key1
-          - key2
+          # Configure list of attribute key patterns that are retained in the resulting stream(s).
```
codeboten, line 252:
```suggestion
          # Configure list of attribute keys that are retained in the resulting stream(s).
```

examples/kitchen-sink.yaml
codeboten: Done

jack-berg (approved):

PR #110, opened 2024-08-06
[SCHEMA] The MetricExporter schema is unsafe

marcalff:
Fixes #109

## Changes

Split `MetricExporter` into:

* `PushMetricExporter`, which can represent `otlp`, `console`, or an extension point,
* `PullMetricExporter`, which can represent `prometheus`, or an extension point,

This is to enforce that:

* `PeriodicMetricReader` can only use a `PushMetricExporter`
* `PullMetricReader` can only use a `PullMetricExporter`

PR #115, opened 2024-08-30
Prefer arrays of name value pairs over objects

### LINKED ISSUE #72: Improve key/value pairs in headers
In the `Headers` node, key/values are represented like this currently:

```
headers:
  key_a: value_a
  key_b: value_b
  api-key: 1234
```

The key name appear as a node name, on the left hand side (LHS).

Issue 1: This most likely adds some constraints on the name itself, due to yaml.

Can key names that include special characters be represented ?

Issue 2: The key name can not be substituted with environment variables.

This is a concern for keys like `api-key`, where the exact key name to use can be vendor dependent: it is highly desirable to abstract this away.

Proposed solution:

```
headers:
  - key:
    name: key_a
    value: value_a
  - key:
    name: "key_b#from%Hell$$" <-- fixes issue 1
    value: value_b
  - key:
    name: ${VENDOR_AUTH_KEY_NAME} <-- fixes issue 2
    value: ${VENDOR_AUTH_KEY_TOKEN}
```

jack-berg:
Resolves #72.

May also resolve https://github.com/open-telemetry/opentelemetry-specification/issues/2857.

This PR migrates `.resource.attributes` and `*.otlp.headers` to be key-value pairs instead of objects, and adds a modeling rule to follow this pattern for similar situations in the future. To repeat the reasoning from the modeling rule:

> When a type requires a configurable list of name-value pairs (i.e. resource attributes, HTTP headers), model using an array of objects, each with `name` and `value` properties. While an arrays of name-value objects is slightly more verbose an object where each key-value is an entry, it is preferred because:
> * Avoids user input as keys, which ensures conformity with the [snake_case properties](#property-name-case) rule.
> * Allows both the names and the values to be targets for [env var substitution].

The more I think about it, the more I think this the way the way to go. Note that this is how environment variables are specified in kubernetes manifests, so it shouldn't be particularly foreign to users.

We should get this change in sooner than later to minimize the number of users we expose to churn.


examples/kitchen-sink.yaml
```diff
@@ -416,7 +419,23 @@ resource:
...
     #
     # Environment variable: OTEL_SERVICE_NAME
-    service.name: !!str "unknown_service"
+    - name: service-name
+      string_value: "unknown_service"
+    # Configure other resource attributes with explicit types.
+    - name: bool_key
+      bool_value: true
+    - name: int_key
+      int_value: 1
```
trask, line 428: I kind of the modeling below since it simplifies the common case:
```suggestion
    - name: service-name
      value: "unknown_service"
    # Configure other resource attributes with explicit types.
    - name: bool_key
      value: true
      type: bool
    - name: int_key
      value: 1
      type: int
```

jack-berg: That's a nice way to model it 👍

marcalff: I prefer (name, value, type) as well.

Schema wise, not sure if it will be possible to represent a bool value as just `true`, or if it needs to always be a string as `"true"`, with the config.yaml implementation having to convert a string to the proper type, but even then, a triplet (name, value, type) is better.

To address with usage examples.

jack-berg: > Schema wise, not sure if it will be possible to represent a bool value as just true, or if it needs to always be a string as "true", with the config.yaml implementation having to convert a string to the proper type, but even then, a triplet (name, value, type) is better.

And what about arrays of strings, arrays of bools, arrays of ints, arrays of doubles? Same question applies. We could use something like [oneOf](https://json-schema.org/understanding-json-schema/reference/combining#oneOf) and list the allowed types, or treat all values as a string and require arrays to be comma separated and require implementations to parse the string value to the specified `type`.

README.md
```diff
@@ -70,6 +68,24 @@ Properties should be modeled using the most appropriate data structures and type
...
+    ```yaml
+   tracer_provider:
+     processors:
+       - batch:
+           exporter:
+             otlp:
+               headers:
+                - name: ${AUTHORIZATION_HEADER_NAME:-api-key}
+                  value: ${AUTHORIZATION_HEADER_VALUE}
+    ```
```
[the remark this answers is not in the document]
jack-berg: Yes I agree. I punted on that in #106 to minimize the scope of the PR, but definitely think we should add support for `*.otlp.headers_list` to allow compatibility with the `OTEL_EXPORTER_OTLP_{SIGNAL}_HEADERS` env vars.

examples/kitchen-sink.yaml
jack-berg: FYI, I pushed a commit which sets the `.resource.attributes[]` items schema to be:
```
{
  "title": "AttributeNameValue",
  "type": "object",
  "properties": {
      "name": {
          "type": "string"
      },
      "value": {
          "oneOf": [
              {"type": "string"},
              {"type": "number"},
              {"type": "boolean"},
              {"type": "null"},
              {"type": "array", "items": {"type": "string"}},
              {"type": "array", "items": {"type": "boolean"}},
              {"type": "array", "items": {"type": "number"}}
          ]
      },
      "type": {
          "enum": [
              null,
              "string",
              "bool",
              "int",
              "double",
              "string_array",
              "bool_array",
              "int_array",
              "double_array"
          ]
      }
  },
  "required": [
      "name", "value"
  ]
}
```
Notes:
- `.type` can be null, or explicitly set to one of the enumerated values corresponding to the allowed types for resource attributes. 
- If `.type` null, it defaults to be `string`. This behavior is conveyed to users / implementers via a comment in the `kitchen-sink.yaml` example. Note, with complicated semantics like this, #104 is important since it helps ensure consistent comments across all the example files.
- `.value` must be "oneOf" the allowed types. Its up to the implementation to interpret it as the specified `.type`. For example, in the java implementation this "oneOf" results in `.value` being a generic `Object`. We'll need to cast `Object` to the corresponding `.type`, and fail if it can't be cast. 

This all feels a little more brittle and difficult to implement than the first commit in this PR, where we have separate properties for each of the types (i.e. `bool_value`, `string_value`, `int_value`, `double_value`, etc). But it does feel a little better for the simple case.

README.md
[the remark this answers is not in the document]
marcalff: Marking as resolved, to be fixed by #117 independently of this change.

examples/kitchen-sink.yaml
marcalff: I agree the schema is more complex.

Some general observation:

* the yaml schema is written once, by maintainers of this repository.
* the yaml schema will be implemented a dozen times, in each language SDK.
* the yaml config files will be written many, many times, by many people. At scale, it means for every deployment of opentelemetry.

Each time we make the config.yaml more natural for end users, it is a win that will increase adoption and contribute to the overall success of this project, even at the cost of some pain in the schema or for language maintainers (who have to support the schema).

An example of this is #110, with more complexity in the schema, and less chances for users to have invalid configs.

README.md
```diff
@@ -70,6 +68,24 @@ Properties should be modeled using the most appropriate data structures and type
 
 In instances where there is a type mismatch between the JSON schema and equivalent standard env var, an alternative version of the property may be provided to resolve the mismatch. For example, resource attributes are configured at `.resource.attributes`, but `.resource.attributes_list` is available with a format matching that of `OTEL_RESOURCE_ATTRIBUTES`. Alternative properties are reserved for cases where there is a demonstrated need for platforms to be able to participate in configuration and there is no reasonable alternative.
 
+### Name-value pairs
+
+When a type requires a configurable list of name-value pairs (i.e. resource attributes, HTTP headers), model using an array of objects, each with `name` and `value` properties. While an arrays of name-value objects is slightly more verbose an object where each key-value is an entry, it is preferred because:
```
codeboten, line 73:
```suggestion
When a type requires a configurable list of name-value pairs (i.e. resource attributes, HTTP headers), model using an array of objects, each with `name` and `value` properties. While an array of name-value objects is slightly more verbose than an object where each key-value is an entry, the latter is preferred because:
```

examples/kitchen-sink.yaml
```diff
@@ -411,12 +414,40 @@ tracer_provider:
...
   #
+  # Each entry must contain .name and .value, and may optionally include .type, which defaults to "string" if not set. The value must match the type. Known types include: string, bool, int, double, string_array, bool_array, int_array, double_array.
+  #
   # Environment variable: OTEL_RESOURCE_ATTRIBUTES
   attributes:
     # Configure `service.name` resource attribute
     #
     # Environment variable: OTEL_SERVICE_NAME
-    service.name: !!str "unknown_service"
+    - name: service-name
```
codeboten, line 424:
```suggestion
    - name: service.name
```

codeboten (commented):
Just a couple of comments, otherwise this looks ok to me

examples/kitchen-sink.yaml
jack-berg: @marcalff I agree and think that the current state of this PR is the best path forward.

jack-berg:
@marcalff, @codeboten - I've resolved merge conflicts and incorporated feedback. Can you take another look? Thanks.

codeboten (approved):

PR #116, opened 2024-08-30
Fix build

jack-berg:
@tsloughter it looks like the build is broken after merging #94. Running `make` currently yields and doesn't run all the tasks:
```
cp -R /Users/jberg/code/open-telemetry/opentelemetry-configuration/validator/../schema /Users/jberg/code/open-telemetry/opentelemetry-configuration/validator/
``` 
Any idea what's up?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@ SCHEMA_FILES := $(shell find . -path './schema/*.json' -exec basename {} \; | so
 EXAMPLE_FILES := $(shell find . -path './examples/*.yaml' -exec basename {} \; | sort)
 $(shell mkdir -p out)
 
-include validator/Makefile
-
 .PHONY: all
 all: install-tools compile-schema validate-examples
 
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ $(shell mkdir -p out)
 .PHONY: all
 all: install-tools compile-schema validate-examples
 
+include validator/Makefile
+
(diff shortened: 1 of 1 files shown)


jack-berg:
There's some other things failing as well:
- https://github.com/open-telemetry/opentelemetry-configuration/actions/runs/10638140808/job/29493474553?pr=115
- https://github.com/open-telemetry/opentelemetry-configuration/actions/runs/10638140806/job/29493474573?pr=115

Makefile
```diff
@@ -2,8 +2,6 @@ SCHEMA_FILES := $(shell find . -path './schema/*.json' -exec basename {} \; | so
 EXAMPLE_FILES := $(shell find . -path './examples/*.yaml' -exec basename {} \; | sort)
 $(shell mkdir -p out)
 
-include validator/Makefile
-
 .PHONY: all
 all: install-tools compile-schema validate-examples
```
tsloughter, line 7: Instead do:

```

.PHONY: all
all: install-tools compile-schema validate-examples

include validator/Makefile
```

tsloughter (approved):

tsloughter:
Why didn't this cause the PR checks to fail?

marcalff (approved):
Thanks for the cleanup.

jack-berg:
> Why didn't this cause the PR checks to fail?

The build never calls `make` by itself - it always calls some more specific task.

Issue #112, opened 2024-08-19
Should attribute_keys in stream configuration support exclude lists?


codeboten:
This is a follow up to #98, specifically around adding support for an exclude list. The original issue will be fixed by adding support for include/exclude options to the attribute_keys list.

This issue will remain open until a change is approved or not in the specification

PR #118, opened 2024-09-09
stream: update stream configuration to include/exclude

codeboten:
This follows https://github.com/open-telemetry/opentelemetry-specification/pull/4188 to add support to exclude attribute keys from stream configuration.

Fixes https://github.com/open-telemetry/opentelemetry-configuration/issues/112

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -254,6 +254,8 @@ meter_provider:
           included:
             - key1
             - key2
+          excluded:
+            - key3
 
 # Configure text map context propagators.
 #
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -143,7 +143,7 @@ meter_provider:
               # Configure resource attributes to be included, in this example attributes starting with service.
               included:
                 - "service*"
-              # Configure resource attributes to be excluded, in this example attribute service.attr1.
+              # Configure resource attributes to be excluded, in this example attribute service.attr1. Applies after .with_resource_constant_labels.included (i.e. excluded has higher priority than included).
               excluded:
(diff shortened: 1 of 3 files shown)


codeboten:
Will mark ready for review once the spec change is merged

codeboten:
spec has merged, this is now ready for review

examples/kitchen-sink.yaml
```diff
@@ -262,6 +262,8 @@ meter_provider:
           included:
             - key1
             - key2
+          excluded:
```
jack-berg, line 265: Can you add a comment for `excluded`, and describe the priority between `included` and `excluded`?

codeboten: Done!

examples/kitchen-sink.yaml
```diff
@@ -262,6 +262,7 @@ meter_provider:
           included:
             - key1
             - key2
+          # Configure resource attributes to be excluded, in this example attribute service.attr1.
```
jack-berg:
```suggestion
          # Configure resource attributes to be excluded, in this example attribute key3.
```
This brings up a good point. It seems like this needs to be _either_ a list to include, OR a list to exclude, but not both?

codeboten: Does it? I would expect this to be similar to how resource attributes can be included or excluded. The example in the kitchen sink shows an include of `service.*` and an exclude of `service.attr1`

jack-berg (approved):

examples/kitchen-sink.yaml
jack-berg: Not sure how SDKs would handle the presence of both include and exclude lists. I know java (and probably go) can handle.

PR #121, opened 2024-09-16
add log_level configuration

### LINKED ISSUE #120: Add log level property to schema
As indicated in `sdk-migration-config.yaml`, we [don't currently have](https://github.com/open-telemetry/opentelemetry-configuration/blob/05a15a86b2ebda55731fb9e959efd2e2d609353d/examples/sdk-migration-config.yaml#L17C5-L17C19) any property to represent the [OTEL_LOG_LEVEL](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/sdk-environment-variables.md#general-sdk-configuration) environment variable.

Let's add support for this.

codeboten:
This is to support the env variable OTEL_LOG_LEVEL.

Fixes #120

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -12,7 +12,7 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not.
 # If omitted or null, false is used.
 disabled: false
-# Configure the log level used by the SDK logger
+# Configure the log level of the internal logger used by the SDK
 # If omitted, info is used.
 log_level: info
 # Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -12,7 +12,7 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not.
 # If omitted or null, false is used.
 disabled: false
-# Configure the log level of the internal logger used by the SDK
+# Configure the log level of the internal logger used by the SDK.
 # If omitted, info is used.
 log_level: info
 # Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -12,6 +12,9 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not.
 # If omitted or null, false is used.
 disabled: false
+# Configure the log level used by the SDK logger
+# If omitted, info is used.
+log_level: info
 # Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
 attribute_limits:
   # Configure max attribute value size. 
--- a/schema/opentelemetry_configuration.json
+++ b/schema/opentelemetry_configuration.json
@@ -11,6 +11,9 @@
         "disabled": {
             "type": ["boolean", "null"]
         },
+        "log_level": {
+            "type": ["string", "null"]
+        },
         "attribute_limits": {
             "$ref": "#/$defs/AttributeLimits"
         },
--- a/examples/sdk-config.yaml
+++ b/examples/sdk-config.yaml
@@ -12,7 +12,7 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not.
 # If omitted or null, false is used.
 disabled: false
-# Configure the log level of the internal logger used by the SDK
+# Configure the log level of the internal logger used by the SDK.
 # If omitted, info is used.
 log_level: info
 # Configure resource for all signals.
--- a/examples/sdk-config.yaml
+++ b/examples/sdk-config.yaml
@@ -12,7 +12,7 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not.
 # If omitted or null, false is used.
 disabled: false
-# Configure the log level used by the SDK logger
+# Configure the log level of the internal logger used by the SDK
 # If omitted, info is used.
 log_level: info
 # Configure resource for all signals.
--- a/examples/sdk-config.yaml
+++ b/examples/sdk-config.yaml
@@ -12,6 +12,9 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not.
 # If omitted or null, false is used.
 disabled: false
+# Configure the log level used by the SDK logger
+# If omitted, info is used.
+log_level: info
 # Configure resource for all signals.
 # If omitted, the default resource is used.
 resource:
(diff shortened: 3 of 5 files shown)


examples/kitchen-sink.yaml
```diff
@@ -15,6 +15,11 @@ file_format: "0.1"
 # Environment variable: OTEL_SDK_DISABLED
 disabled: false
 
+# Configure the log level used by the SDK logger
+#
+# Environment variable: OTEL_LOG_LEVEL
+log_level: info
+
```
trask, line 22: do you think it makes sense to add a top-level `sdk` node in order to scope the SDK configuration now that we also support `instrumentation` config?

codeboten: I kinda think of the root of the doc as the SDK's node. I could be convinced otherwise if folks feel strongly. I'd suggest discussing in a separate issue

marcalff:
Related:

* https://github.com/open-telemetry/opentelemetry-specification/issues/2039

Until the spec defines an enum for log levels, opentelemetry-configuration can only represent a log level as a string, to be interpreted (differently) by each SDK implementation.

examples/kitchen-sink.yaml
marcalff: `disabled` is an SDK property, and at the top level, so this is consistent as is.
LGTM.

marcalff (approved):
LGTM.

This is the best we can do now, because the spec does not define an enum for log levels.

examples/kitchen-sink.yaml
```diff
@@ -15,6 +15,11 @@ file_format: "0.1"
 # Environment variable: OTEL_SDK_DISABLED
 disabled: false
 
+# Configure the log level used by the SDK logger
```
jack-berg, line 18: What is the SDK logger? I think you mean to say the internal logger used internally by the SDK.

schema/opentelemetry_configuration.json
```diff
@@ -11,6 +11,9 @@
         "disabled": {
             "type": ["boolean", "null"]
         },
+        "log_level": {
+            "type": ["string", "null"]
```
jack-berg, line 15: Can we have an enum based on a subset of the SeverityNumber enum defined in the proto repo? https://github.com/open-telemetry/opentelemetry-proto/blob/4f69356d853029975649c3f38b06fc77d77975fc/opentelemetry/proto/logs/v1/logs.proto#L86-L110

examples/kitchen-sink.yaml
codeboten: Copied from 
https://github.com/open-telemetry/opentelemetry-specification/blob/d6977d20eb22b3467fc5ddab43ac94bffa38ccea/specification/configuration/sdk-environment-variables.md?plain=1#L117

jack-berg: I know but I think that definition is lame. Can / should we do better?

codeboten, line 18:
```suggestion
# Configure the log level of the internal logger used by the SDK
```

codeboten: Updated, PTAL

schema/opentelemetry_configuration.json
codeboten: We could, though without a resolution on https://github.com/open-telemetry/opentelemetry-specification/issues/2039, we may be jumping ahead of the spec

jack-berg (commented):
We don't use the `OTEL_LOG_LEVEL` property in `opentelemetry-java`. For internal SDK logging, we use java util logging (or JUL for short), which comes with a default log level already, and has standard options for changing the log level on a per-library basis. For us, introducing an additional option would only muddy the log configuration story for users.

Therefore, I'm not a good judge of this property of its value since to me from the perspective of a java maintainer, it seems like a mistake. I won't approve, but also won't block progress.

examples/kitchen-sink.yaml
```diff
@@ -11,6 +11,11 @@ file_format: "0.3"
 # Configure if the SDK is disabled or not. This is not required to be provided to ensure the SDK isn't disabled, the default value when this is not provided is for the SDK to be enabled.
 disabled: false
 
+# Configure the log level of the internal logger used by the SDK
+#
+# Environment variable: OTEL_LOG_LEVEL
+log_level: info
```
jack-berg, line 17: Please add the property description to [type_descriptions.yaml](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/type_descriptions.yaml) to ensure consistency across the examples.

codeboten:
> For internal SDK logging, we use java util logging (or JUL for short), which comes with a default log level already, and has standard options for changing the log level on a per-library basis. For us, introducing an additional option would only muddy the log configuration story for users.

I would expect this property to be used in configuring JUL in the java implementation, would that not be the case? Is there an alternative configuration property available to end users through the configuration file?

examples/kitchen-sink.yaml
codeboten: Done 👍🏻

jack-berg:
> I would expect this property to be used in configuring JUL in the java implementation, would that not be the case? Is there an alternative configuration property available to end users through the configuration file?

I mean we'd have to have a discussion about it in the java SIG but my given that we've gone this long without supporting `OTEL_LOG_LEVEL`, I believe we'd skip support for this property under the same reasoning.

> Is there an alternative configuration property available to end users through the configuration file?

No. Different tools for different things. The logging ecosystem in java is fractured and to complicated to explain here, but essentially users are used to the idea that they configure the logs from their libraries / frameworks in particular way. If we add support for this property, then there's the question of priority - should / can this property override the user's separate log config? A [search](https://github.com/open-telemetry/opentelemetry-java/issues?q=is%3Aissue+OTEL_LOG_LEVEL+is%3Aclosed) for `OTEL_LOG_LEVEL` in `opentelemetry-java` doesn't indicate that there's been much discussion on this topic, the opinion I'm expressing here is worth revisiting. But its also noteworthy that we're this far into the  of `opentelemetry-java`, and haven't heard much demand for `OTEL_LOG_LEVEL`.

codeboten:
A quick search through the OpenTelemetry github org https://github.com/search?q=org%3Aopen-telemetry%20OTEL_LOG_LEVEL&type=code shows that it's used in a few different SIGs, so it probably makes sense to move forward with it.

@brettmc can you review as it looks like PHP is one of the users of this env variable

brettmc:
> can you review as it looks like PHP is one of the users of this env variable

@codeboten sorry I'm late with the review, but the proposed usage would work for PHP's requirements for log level. :+1:

PR #141, opened 2024-11-25
Provide guidance on required and null properties

jack-berg:
I analyzed JSON schema's `required` and `type: ["null"]` concepts and have provided guidance on when and how to use them in schema modeling. Also, I've carefully applied this guidance to existing types. 

I've been thinking we've needed this for a while, but this conversation was the impetus: https://github.com/open-telemetry/opentelemetry-configuration/pull/140#issuecomment-2498411340

### DIFF
--- a/schema/meter_provider.json
+++ b/schema/meter_provider.json
@@ -2,7 +2,7 @@
     "$id": "https://opentelemetry.io/otelconfig/meter_provider.json",
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "title": "MeterProvider",
-    "type": ["object", "null"],
+    "type": "object",
     "additionalProperties": false,
     "properties": {
         "readers": {
@@ -26,9 +26,12 @@
             ]
         }
     },
+    "required": [
+        "readers"
+    ],
     "$defs": {
         "PeriodicMetricReader": {
-            "type": ["object", "null"],
+            "type": "object",
             "additionalProperties": false,
             "properties": {
                 "interval": {
@@ -49,7 +52,7 @@
             "title": "PeriodicMetricReader"
         },
         "PullMetricReader": {
-            "type": ["object", "null"],
+            "type": "object",
             "additionalProperties": false,
             "properties": {
                 "exporter": {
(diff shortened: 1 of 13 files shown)


schema/meter_provider.json
```diff
@@ -2,7 +2,7 @@
     "$id": "https://opentelemetry.io/otelconfig/meter_provider.json",
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "title": "MeterProvider",
-    "type": ["object", "null"],
+    "type": "object",
```
jack-berg, line 5: Setting `meter_provider:` is not a sensible thing to do. If you want a noop meter_provider, omit `meter_provider` altogether.

schema/meter_provider.json
```diff
@@ -26,9 +26,12 @@
             ]
         }
     },
+    "required": [
+        "readers"
```
jack-berg, line 30: A `meter_provider` without readers isn't sensible.

schema/tracer_provider.json
```diff
@@ -211,9 +214,6 @@
                     "minimum": 0
                 }
             },
-            "required": [
-                "endpoint"
```
jack-berg, line 215: Endpoint is not actually required for zipkin since there's a well defined default value.

schema/type_descriptions.yaml
```diff
@@ -54,30 +79,39 @@
...
+      
+      If omitted or null, 128 is used.
   path_patterns:
     - .attribute_limits
 
 - type: Propagator
   property_descriptions:
-    composite: Configure the set of propagators to include in the composite text map propagator.
+    composite: >
+      Configure the set of propagators to include in the composite text map propagator. Values include: tracecontext, baggage, b3, b3multi, jaeger, xray, ottrace, none. For behavior of values see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration.
```
marcalff, line 114: In the spec:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#propagators-distribution

`xray` is not listed in "The official list of propagators"

Maybe clarify that propagators can be:

* from the list of official propagators (builtins)
* from extension points

Also, a propagator of `none` is dubious, this is yet another way to specify null with a different syntax.
There should be no `propagator:` node in this case.

marcalff (approved):
LGTM, thanks for the fix.

PR #142, opened 2024-11-26
Add a variety of key definitions surrounding stability

### LINKED ISSUE #89: Define stability guarantees for stable releases

jack-berg:
Related to #2.

Resolves #89.

Took inspiration from prior art in the [opentelemetry-proto stability definition](https://github.com/open-telemetry/opentelemetry-proto?tab=readme-ov-file#stability-definition), but adapted for a variety of differences in the the domains of OTLP vs. declarative configuration, protobuf vs. JSON schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,41 @@ When a type requires a configurable list of name-value pairs (i.e. resource attr
                   value: ${AUTHORIZATION_HEADER_VALUE}
     ```
 
+### Required and null properties
+
+JSON schema has two related but subtly different concepts involved in indicating the requirement level of properties and values:
+
+* [`type` of `null`](https://json-schema.org/understanding-json-schema/reference/null): When a property includes a type of `null` along with other allowed types (i.e. `"type": ["string", "null"]`), it indicates that even if the property key is present, the value may be omitted. This is useful in a variety of situations:
+  * When modeling properties with primitive types which are candidates for [env var substitution][], since allowing `null` means that the configuration is valid even if the referenced env var is undefined.
+  * When modeling objects which do not require any properties. In these cases, either no properties are required, or there are no properties and the presence of the property key expresses the desired state.
+* [required](https://json-schema.org/understanding-json-schema/reference/object#required): When a property is `required`, the key must be included in the object or the configuration is invalid. Properties should be required when there is no well default semantic (i.e. it's not clear what the behavior is when the property is absent).
+
+For example:
+
+```
+tracer_provider:
+ processors:
+   - simple:
+       exporter:
+         console:
+ limits:
+   attribute_value_length_limit: ${OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT}
+```
+
+* `tracer_provider` is not required. When omitted, a noop tracer provider is used.
+* `tracer_provider`'s type is `object`. There's no sensible tracer provider which does not minimally set one entry in `processors`. 
+* `exporter` is required. A simple processor without an exporter is invalid.
+* `exporter`'s type is `object`. Setting `exporter` to `null` or any non-object value is invalid.
+* `console`'s type is `["object", "null"]`. The console exporter has no properties, and we should not force the user to set an empty object (i.e `console: {}`).
+* `limits` is not required. When omitted, default span limits are used.
+* `limits`'s type is `object`. If a user includes the `limits` property, they must set at least one property. Settings `limits` to `null` is invalid.
+* `attributes_value_length_limit` is not required. If omitted, no attribute length limits are applied.
+* `attributes_value_length_limit`'s type is `["integer", "null]`. If null (i.e. because the `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` env var is unset), no attribute length limits are applied.
+
+If a property is _not_ required, it should include a [comment](./CONTRIBUTING.md#description-generation) describing the semantics when it is omitted.
+
+If a property `type` includes `null`, it must include a [comment](./CONTRIBUTING.md#description-generation) describing the semantics when the value is `null`. It's common for properties with primitive types to allow `null`. `object` types allow `null` if no properties are required and the presence of the property key is meaningful. 
+
 ## Contributing
 
 See [CONTRIBUTING.md](CONTRIBUTING.md)
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Stable types provide the following guarantees. All types except those excluded i
   * [required](https://json-schema.org/understanding-json-schema/reference/object#required): will not add additional entries.
   * [contains, minContains, maxContains](https://json-schema.org/understanding-json-schema/reference/array#contains): will not become stricter.
   * [minItems, maxItems](https://json-schema.org/understanding-json-schema/reference/array#length): `minItems` will not increase, `maxItems` will not decrease.
-  * [uniqueItems](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems): will not go from `true` to `false`.
+  * [uniqueItems](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems): will not go from `false` to `true`.
   * [enum](https://json-schema.org/understanding-json-schema/reference/enum): will not remove entries.
   * [const](https://json-schema.org/understanding-json-schema/reference/const): will not change.
 * No existing type will be deleted.
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ The following additive changes are allowed:
 * Adding of new properties to existing types.
 * Adding new types.
 * Changes that make property validation less strict. See above for examples.
+* Removing a property from `required`.
 
 ### Applicability
 
@@ -104,7 +105,7 @@ The top level `.file_format` property is special in that it conveys the version 
 Given the [guarantees and allowed changes](#guarantees-and-allowed-changes), implementations may encounter the following scenarios:
 
 * The `file_format` major version aligns with the implementation major version, AND:
-  * The `file_format` minor version is less than or equal to the implementation minor version: This is idea, with versions maximally aligned. Despite this, an implementation might not support every property and type of its target version.
+  * The `file_format` minor version is less than or equal to the implementation minor version: This is ideal, with versions maximally aligned. Despite this, an implementation might not support every property and type of its target version.
   * The `file_format` minor version is greater than the implementation minor version: The implementation should detect and emit a warning since there may be configuration features the user specifies which the implementation does not understand. However, this is acceptable in many cases, and not terribly different from the ideal path where an implementation also might not support every configuration feature.
 * The `file_format` major version does not align with the implementation major version. The implementation should produce an error, since there may be breaking changes in the properties and semantics on how they are interpreted. Implementations may choose to temporarily support multiple major version to accommodate transitioning users.
 
--- a/README.md
+++ b/README.md
@@ -109,89 +109,6 @@ Given the [guarantees and allowed changes](#guarantees-and-allowed-changes), imp
   * The `file_format` minor version is greater than the implementation minor version: The implementation should detect and emit a warning since there may be configuration features the user specifies which the implementation does not understand. However, this is acceptable in many cases, and not terribly different from the ideal path where an implementation also might not support every configuration feature.
 * The `file_format` major version does not align with the implementation major version. The implementation should produce an error, since there may be breaking changes in the properties and semantics on how they are interpreted. Implementations may choose to temporarily support multiple major version to accommodate transitioning users.
 
-## Schema modeling rules
-
-The following rules are enforced when modeling the configuration schema:
-
-### What properties are part of schema?
-
-Only properties which are described in [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification) or [semantic-conventions](https://github.com/open-telemetry/semantic-conventions) are modeled in the schema. However, it's acceptable to allow additional properties specific to a particular language or implementation, and not covered by the schema. Model these by setting `"additionalProperties": true` (see [JSON schema additionalProperties](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties)). Types should set `"additionalProperties": false` by default unless requested by an opentelemetry component [maintainer](https://github.com/open-telemetry/community/blob/main/community-membership.md#maintainer) which supports additional options.
-
-### Property naming
-
-To remove redundant information from the configuration file, prefixes for data produced by each of the providers will be removed from configuration options. For example, under the `meter_provider` configuration, metric readers are identified by the word `readers` rather than by `metric_readers`. Similarly, the prefix `span_` will be dropped for tracer provider configuration, and `logrecord` for logger provider.
-
-### Property name case
-
-Properties defined in the schema should be lower [snake case](https://en.wikipedia.org/wiki/Snake_case).
-
-### Properties which pattern matching
-
-When a property requires pattern matching, use wildcard `*` (match any number of any character, including none) and `?` (match any single character) instead of regex. If a single property with wildcards is likely to be insufficient to model the configuration requirements, accept `included` and `excluded` properties, each with an array of strings with wildcard entries. The wildcard entries should be joined with a logical OR. If `included` is not specified, assume that all entries are included. Apply `excluded` after applying `included`. Examples:
-
-* Given `excluded: ["a*"]`: Match all except values starting with `a`.
-* Given `included: ["a*", "b*"]`, `excluded: ["ab*"]`: Match any value starting with `a` or `b`, excluding values starting with `ab`.
-* Given `included: ["a", "b"]`, `excluded: ["a"]`: Match values equal to `b`.
-
-### Data modeling and environment variable substitution
-
-Properties should be modeled using the most appropriate data structures and types to represent the information. This may result in a schema which doesn't support env var substitution for the [standard env vars](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md) where a type mismatch occurs. For example, the `OTEL_RESOURCE_ATTRIBUTES` env var is modeled as a string, consisting of a comma separated list of key-value pairs, which is not the natural way to model a mapping of key-value pairs in JSON schema. 
-
-In instances where there is a type mismatch between the JSON schema and equivalent standard env var, an alternative version of the property may be provided to resolve the mismatch. For example, resource attributes are configured at `.resource.attributes`, but `.resource.attributes_list` is available with a format matching that of `OTEL_RESOURCE_ATTRIBUTES`. Alternative properties are reserved for cases where there is a demonstrated need for platforms to be able to participate in configuration and there is no reasonable alternative.
-
-### Name-value pairs
-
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -28,15 +28,85 @@ go-jsonschema \
...
+
+A secondary objective is to allow for stable code generation from the JSON schema for language implementations of the [in-memory configuration model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#in-memory-configuration-model). With stability guarantees in mind, maintainers should be able to write code generation logic (or handcraft an in-memory configuration model) without risk that allowed changes to the schema will break stable implementations. **NOTE**: There is _no_ guarantee that the output of off-the-shelf [code generation](#code-generation) tools will be stable when allowed changes are made. Maintainers are responsible for understanding how code generation tools work (and evolve) and intersect with the guarantees made here.
+
+All schema changes are considered through the lens of maintaining these objectives.
+
+### Guarantees and allowed changes
```
marcalff, line 48: This whole section is written as "absolute" terms.

In reality, every statement here can be invalid, when interpreted in a different, "relative" context.

By absolute/relative, I mean this:

The format is not just a contract between two (2) parties.

The format is a contract between three (3) parties, which are:

* [SCHEMA] maintainers of the opentelemetry-configuration repository, when defining a yaml schema
* [SIG] maintainers of a SIG, when parsing configuration from a yaml schema
* [USER] maintainers of a given user configuration files, when writing a yaml schema configuration

I think it will be beneficial to details separately:

* the contract and expectations between [SCHEMA] and [SIG]
* the contract and expectations between [SCHEMA] and [USER]
* the contract and expectations between [USER] and [SIG]

A claim as "Type property names will not change" is not absolute, some type names will change.

For a given schema version number, the type property name will not change even on minor file_format changes, sure, so this is true in the context of [USER] - [SIG], within some constraints on the file_format value.

This does not forbid name changes when evolving the schema definitions:

* new schema can rename properties, affecting the [SCHEMA] - [SIG] part when parsing, for example on a major version change in file_format
* new schema can rename properties, affecting the [SCHEMA] - [USER] part when writing a yaml config, also on major version change in file_format: users __changing__ the file_format tag in their file are expected to adjust to the new format spec.

jack-berg: > The format is a contract between three (3) parties, which are:

I'm not sure I completely understand what you mean by this. I think part of what you're saying is reflected in the [objectives](https://github.com/jack-berg/opentelemetry-configuration/tree/stability-definition?tab=readme-ov-file#objectives) section, which explains that stability guarantees impact both a user writing YAML config (i.e. [SCHEMA] and [USER] in your terms), and implementations interpretting this config and publishing in-memory representations of the data model (i.e. [SCHEMA] and [SIG] in your terms). 

> A claim as "Type property names will not change" is not absolute, some type names will change.

What do you mean by this? What I mean to claim here is that a given property name (let's say the `.tracer_provider` property for example) which is part of a stable type (i.e. no `*/experimental` suffix) will not change in minor or path versions. 

> new schema can rename properties, affecting the [SCHEMA] - [SIG] part when parsing, for example on a major version change in file_format

Yes, all changes are allowed in major version updates. Do we need to explicitly state this? I'd like to follow the lead of `opentelemetry-proto`'s [stability definitions](https://github.com/open-telemetry/opentelemetry-proto?tab=readme-ov-file#stability-definition) and try to keep language as terse as is required by the language. 

Do you think that it would help to explain that these stability requirements apply to minor version updates, and explain that major version updates can include carte blanche breaking changes.

README.md
```diff
@@ -28,15 +28,85 @@ go-jsonschema \
...
+
+### File format
+
+The top level `.file_format` property is special in that it conveys the version of the schema a user's configuration conforms to. Implementations also target a particular version of the schema, which may or may not align with the version specified by the user.
+
+Given the [guarantees and allowed changes](#guarantees-and-allowed-changes), implementations may encounter the following scenarios:
+
+* The `file_format` major version aligns with the implementation major version, AND:
+  * The `file_format` minor version is less than or equal to the implementation minor version: This is idea, with versions maximally aligned. Despite this, an implementation might not support every property and type of its target version.
```
marcalff, line 107: Typo: This is `ideal` ?

README.md
```diff
@@ -28,15 +28,85 @@ go-jsonschema \
...
+  * [propertyNames](https://json-schema.org/understanding-json-schema/reference/object#propertyNames): will not become stricter.
+  * [minProperties, maxProperties](https://json-schema.org/understanding-json-schema/reference/object#size): `minProperties` will not increase, `maxProperties` will not decrease.
+  * [required](https://json-schema.org/understanding-json-schema/reference/object#required): will not add additional entries.
+  * [contains, minContains, maxContains](https://json-schema.org/understanding-json-schema/reference/array#contains): will not become stricter.
+  * [minItems, maxItems](https://json-schema.org/understanding-json-schema/reference/array#length): `minItems` will not increase, `maxItems` will not decrease.
+  * [uniqueItems](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems): will not go from `true` to `false`.
+  * [enum](https://json-schema.org/understanding-json-schema/reference/enum): will not remove entries.
+  * [const](https://json-schema.org/understanding-json-schema/reference/const): will not change.
+* No existing type will be deleted.
```
marcalff, line 73: Assume a format 1.0.0 (stable) that contains the `Jaeger` exporter, which we all know has been deprecated and removed from the spec.

What this says is that the jeager exporter will never be removed from the yaml format in version 1.x.y, as this requires a major format change.

There must be a reasonable way to deprecate and remove features, and I don't think the following is shocking:

- format 1.0.0 does support a jeager exporter
- format 1.1.0 removes support for jaeger, and documents this as an incompatible breaking change

README.md
jack-berg: > A major update in the yaml format will have a massive impact:

I agree the impact is huge.

The impact is so huge that it creates a strong incentive to avoid a major version update. And if the rules are fairly (and necessarily) restrictive in terms of what is allowed in a minor version update, then it creates a strong incentive to model types correctly the first time. 

Consider other areas where similar strict restrictions exist like [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto?tab=readme-ov-file#stability-definition), or the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md), or language implementations like [opentelemetry-java](https://github.com/open-telemetry/opentelemetry-java/blob/main/VERSIONING.md#compatibility-requirements). 

We have definitions of what is allowable in minor versions to protect users, and to promote careful / thoughtful design. In all these cases, a major version bump is something to strongly avoid. We write down when a major version is required with the hope that we don't have to do it. 

For your examples:

1. The first example is the removal of a known property / type, the `jaeger` exporter. This is considered a breaking change under the rules laid out in this PR. Luckily for us, `jaeger` is already gone, but the same could happen to other exporters, processors, samplers, etc. The fact that removing a type is a breaking change is an incentive to keep it some form (potentially deprecated) until the next major version, which could be indefinitely if there is never a strong enough reason. Its not an incentive to make invasive changes and bump the major version. 
2. The second example is the adjustment of constraints from allowing `[0, 100]` to `[10, 90]`. This is considered a breaking change under the rules laid out in this PR. I think this is correct. In the migration path you lay out, the change in behavior is broken into multiple minor versions, but this doesn't change the fact that a value of `5` was valid in 1.0.0, and became invalid in 1.2.0. No major version change occurred yet the user's input became invalid. I think this is wrong, regardless of whether or not it is broken into steps. For authors of the schema, the fact that making validation more strict requires a major version bump is an incentive to get the validation criteria correct up front. The PR lists all the different types of validation conditions present in JSON schema for completeness, but I really don't see us using those in most cases. `minLength`, `maxLength`, `format`, `multipleOf` and others seem like really niche keywords we really ought to avoid. I do anticipate (from experience so far) we'll use keywords like `pattern`, `propertyNames`, `minProperties`, `maxProperties`, `required`, and `enum` frequently. It seems reasonable to model these correctly up front. And if we're not sure, we can start off stricter and loosen the criteria as needed, since reducing strictness is allowed.


> Item (b) is about stability garantees between [SCHEMA] and [SIG] as I called it, where the SIG implementation is to expect extra processing beyond yaml parsing, mandated by the schema.

On a somewhat unrelated note: Right now the extra processing responsibilities of the SIG implementation are sometimes implied and other times made explicit in comments. This seems flimsy, and error prone as SIG maintainers may miss some subtle change in the expected semantics. To me, this implies we should minimize these extra responsibilities.

README.md
jack-berg: > There must be a reasonable way to deprecate and remove features, and I don't think the following is shocking:

I think there must be a way to deprecate features, but not removal. Deprecated features must linger on (potentially indefinitely) until a major version bump. This shifts the burden of understanding / dealing with schema evolution from users to maintainers, which is where it belongs.

Semantic conventions reached this same conclusion: https://opentelemetry.io/docs/specs/otel/versioning-and-stability/#deprecated

I believe (and hope) language SIGs reach this same conclusion. In java, we maintain code which we marked as stable but have deprecated with the same gaurantees as other stable code - no removal until next major version, which we have no plans of. Here's an example. We accidentally exposed a [protected constructor](https://github.com/open-telemetry/opentelemetry-java/blob/0ef86fae7d1c568c3f2dabe6c7b536c54000a2d3/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanLimits.java#L52-L58), which is not aligned with our code style in the repository. Even though its unlikely anyone was / is using this, we've maintained it for 3 years, and will continue maintaining it.

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
+
+### File format
+
+The top level `.file_format` property is special in that it conveys the version of the schema a user's configuration conforms to. Implementations also target a particular version of the schema, which may or may not align with the version specified by the user.
+
+Given the [guarantees and allowed changes](#guarantees-and-allowed-changes), implementations may encounter the following scenarios:
+
+* The `file_format` major version aligns with the implementation major version, AND:
+  * The `file_format` minor version is less than or equal to the implementation minor version: This is ideal, with versions maximally aligned. Despite this, an implementation might not support every property and type of its target version.
+  * The `file_format` minor version is greater than the implementation minor version: The implementation should detect and emit a warning since there may be configuration features the user specifies which the implementation does not understand. However, this is acceptable in many cases, and not terribly different from the ideal path where an implementation also might not support every configuration feature.
```
carlosalberto, line 109: So, as an example: minLength is 10 in 1.1.0 and minLength is 5 for 1.2.0 (hence relaxing the value). User of **1.2.0** expects 7 to be a valid value, but for 1.1.0 it is not. What do you do here? Generate a warning _and_ accept the value? Or ignore it?

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
+  * [additionalProperties](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties): will not go from `true` to `false`.
+  * [propertyNames](https://json-schema.org/understanding-json-schema/reference/object#propertyNames): will not become stricter.
+  * [minProperties, maxProperties](https://json-schema.org/understanding-json-schema/reference/object#size): `minProperties` will not increase, `maxProperties` will not decrease.
+  * [required](https://json-schema.org/understanding-json-schema/reference/object#required): will not add additional entries.
+  * [contains, minContains, maxContains](https://json-schema.org/understanding-json-schema/reference/array#contains): will not become stricter.
+  * [minItems, maxItems](https://json-schema.org/understanding-json-schema/reference/array#length): `minItems` will not increase, `maxItems` will not decrease.
+  * [uniqueItems](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems): will not go from `true` to `false`.
```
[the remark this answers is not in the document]
jack-berg: Yes, good catch

README.md
jack-berg: > User of 1.2.0 expects 7 to be a valid value, but for 1.1.0 it is not. 

So the user has specified that they are using 1.2.0 by setting `file_format: 1.2`. The implementation is conforming to 1.1.0. The implementation detects that the user's version is great than the implementation and emits a warning, alerting the user to the fact that a 1.1.0 implementation won't be able to understand some of the changes from a 1.2.0 file format. 

In this case, the implementation rejects the value as its outside of the acceptable range according to the schema its aware of. If implementations stay up to date with `opentelemetry-configuration` releases, it will be much more common for the implementation's version to be greater than the users (which is ideal) than for the user's version to be greater than the implementations.

codeboten:
This was discussed at 03-Feb-2025 WG call, this needs to be reviewed by the next call (17-Feb)

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
-- For minor versions: TBD
+### Objectives
+
+The primary objective of stability is to protect users from breaking changes. That is, users providing configuration conforming to a particular stable version of the schema should be able to reliably upgrade minor and patch versions without risk that their configuration becomes invalid, or that the interpretation changes.
+
+A secondary objective is to allow for stable code generation from the JSON schema for language implementations of the [in-memory configuration model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#in-memory-configuration-model). With stability guarantees in mind, maintainers should be able to write code generation logic (or handcraft an in-memory configuration model) without risk that allowed changes to the schema will break stable implementations. **NOTE**: There is _no_ guarantee that the output of off-the-shelf [code generation](#code-generation) tools will be stable when allowed changes are made. Maintainers are responsible for understanding how code generation tools work (and evolve) and intersect with the guarantees made here.
```
MrAlias, line 44: This secondary objective seems much like it is already a part of the primary objective. If we do not want to have users upgrade and have their configuration interpretation change we need to ensure the code that is generated, and ultimately creates their SDK, remains consistently stable.

I'm not sure we can make that guarantee though. I seems like it is going to be up to the implementations to ensure this stability when they generate the code.

Is it enough to say we do not plan to break backwards compatibility of the configuration itself and that we will follow the guidelines listed below to ensure implementations do not break. But it is the responsibility of the implementation to comply with their language specific stability requirements?

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
+* [Objectives](#objectives): Overview of the motivation behind stability.
+* [Guarantees and allowed changes](#guarantees-and-allowed-changes): Specific details on allowed and disallowed changed.
+* [Applicability](#applicability): Limits of stability definitions, including experimental features and extension points.
+* [File format](#file-format): The `file_format` property and implementation behavior when schema versions are not aligned.
 
-- For major versions: All changes are permitted.
-- For minor versions: TBD
+### Objectives
+
+The primary objective of stability is to protect users from breaking changes. That is, users providing configuration conforming to a particular stable version of the schema should be able to reliably upgrade minor and patch versions without risk that their configuration becomes invalid, or that the interpretation changes.
```
MrAlias, line 42: This assumes semver versioning. We should specify this, right?

MrAlias (approved):

marcalff:
Resolving my own comments to start a fresh review.

README.md
codeboten, line 44:
```suggestion
A secondary objective is to allow for stable code generation from the JSON schema for language implementations of the [in-memory configuration model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#in-memory-configuration-model). With stability guarantees in mind, maintainers should be able to write code generation logic (or handcraft an in-memory configuration model) without risk that allowed changes to the schema will break stable implementations. **NOTE**: There is _no_ guarantee that the output of off-the-shelf [code generation](#code-generation) tools will be stable when allowed changes are made. Maintainers are responsible for understanding any generated code intersects with the guarantees made here.
```

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
+
+All schema changes are considered through the lens of maintaining these objectives.
+
+### Guarantees and allowed changes
+
+A type is the JSON schema analog to a [protobuf message](https://protobuf.dev/programming-guides/proto3/#simple). Types have a [`type`](https://json-schema.org/understanding-json-schema/reference/type) of `object`, and use various keywords to describe their properties and conditions which constitute valid data.
```
codeboten, line 50:
```suggestion
A [type](https://json-schema.org/understanding-json-schema/reference/type) is the JSON schema analog to a [protobuf message](https://protobuf.dev/programming-guides/proto3/#simple). Types have a `type` of `object`, and use various keywords to describe their properties and conditions which constitute valid data.
```

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
+
+Stable types provide the following guarantees. All types except those excluded in [applicability](#applicability) are considered stable after a 1.0.0 release.
+
+* Type property names will not change.
+* The `type` of properties will not change, except the allowed addition of `null`.
+* Type [title](https://json-schema.org/understanding-json-schema/reference/annotations) will not change.
```
codeboten, line 56: Is the title important to capture here?

README.md
```diff
@@ -28,15 +28,86 @@ go-jsonschema \
...
+  * [const](https://json-schema.org/understanding-json-schema/reference/const): will not change.
+* No existing type will be deleted.
+* No type property will be deleted.
+
+The following additive changes are allowed:
+
+* Adding of new properties to existing types.
+* Adding new types.
+* Changes that make property validation less strict. See above for examples.
+* Removing a property from `required`.
```
codeboten, line 81: Should marking a feature deprecated be captured here? Maybe we can capture this with some form of annotation

README.md
jack-berg: The title may be used by JSON schema code gen tools. Specifying this allows maintainers to make informed decisions about what they can / cannot rely on.

README.md
jack-berg: JSON schema has a concept called [annotations](https://json-schema.org/understanding-json-schema/reference/annotations), with a specific `deprecated` keyword: 

> JSON Schema includes a few [keywords](https://json-schema.org/learn/glossary#keyword), that aren't strictly used for validation, but are used to describe parts of a schema.

The key bit here is that annotations are part of validation. Though (and maybe this is what you're getting at) they may be used in code generation, similar to the `title` annotation.

I'm happy to include a note about deprecated here. I think we should be free to add / remove deprecated freely. I.e. it would be acceptable to add it, then later remove it.

README.md
jack-berg: Yes, good call. I've added this all the way at the top, so that sections can reference `MAJOR`, `MINOR`, `PATCH` where needed.

README.md
jack-berg: I ended up adding bullet points for `description` and `deprecated` annotations.

jack-berg:
Curious if other reviewers @marcalff, @codeboten, @carlosalberto are planning on picking this back up.

In the config SIG, we discussed how it will be important to land something specifying stability requirements so we have a chance to evaluate changes through the context of the requirements and have the chance to iterate / improve. Best case would have had these requirements in place 6-12 months. But next best time is now, while there are still a handful of modeling decisions to be made before stability.

If you don't have time to pick this back up soon, do you feel comfortable with this language as a starting point?

codeboten (approved):
The language looks ok to me, thanks for addressing my comments

jack-berg:
Merging because the "Validator Tests / Push Docker image to Docker Hub" failure is unrelated. Opened #178 to track.

PR #146, opened 2024-12-04
Refactor otlp exporter

### LINKED ISSUE #103: Simplify endpoint
The existing environment variables used in the spec to describe an endpoint:

* OTEL_EXPORTER_OTLP_PROTOCOL
* OTEL_EXPORTER_OTLP_ENDPOINT
* OTEL_EXPORTER_OTLP_INSECURE

are the result of incremental design, due to history.

The current design for config.yaml derives from this existing specification, and currently is:

```
otlp:
  # http/protobuf, http/json, or grpc
  protocol:

  # http endpoint, https endpoint, or no scheme
  endpoint:

  # true or false, only used for grpc protocol with no scheme endpoint
  insecure: 
```

Exposing the same fields to match the existing environment variables has some merits, but also carries over existing issues.

# Existing configuration

1. Protocol

The protocol to use should be in the endpoint URL.
Instead, when using grpc, the "protocol" is set to "grpc", while the endpoint may contain "http" or "https".
This makes no sense.

2. Message format

The message format, "protobuf" or "json", is mixed with the protocol.
I don't think it belongs there.

3. Security

The "insecure" field is a last resort configuration parameter for grpc when not used with http/https.
This is too convoluted.

4. Lack or orthogonality

With the existing configuration, if someone changes the protocol between http and grpc, then the following fields must be adjusted as well:

* endpoint needs to be fixed for the port number
* endpoint needs to be fixed to add or remove the /v1/{signal} path
* the endpoint http/https scheme might be missing when coming from grpc, and needs to be fixed
* the insecure field may be used or not, adding general confusion. `insecure=false` may still be insecure.

# Proposed configuration

```
otlp:
  # Valid values are "protobuf" or "json".
  content_type: protobuf

  # Full URL with:
  # - a scheme ("http", "https", "grpc", "grpcs")
  # - a host name
  # - a port number
  # - an optional path, for http
  endpoint:
```

Examples of configuration:

```
otlp:
  content_type: protobuf
  endpoint: "http://localhost:4318/v1/traces"
```

```
otlp:
  content_type: json
  endpoint: "https://localhost:4318/v1/traces"
```

```
otlp:
  content_type: protobuf
  endpoint: "grpc://localhost:4317"
```

```
otlp:
  content_type: protobuf
  endpoint: "grpcs://localhost:4317"
```

Note how this simplifies things, as the endpoint URL specifies:

* the protocol to use, HTTP versus gRPC
* if TLS is used, with http / grpc versus https / grpcs

all this in the same field.

It then becomes much easier to change from:

```
  endpoint: "grpc://localhost:4317"
```

to

```
  endpoint: "https://localhost:4318/v1/traces"
```

when editing a config file, or when providing substitution variables.

# General considerations

Note that I am **NOT** asking to change existing environment variables in the specs, as this will have some global impact on all SIG and all existing deployments.

This ship has sailed.

What I am asking to consider, is to have a different design for config.yaml only, which I think makes more sense and will be more practical to work with in the long term.

Given that there are no existing deployments using config.yaml, there is no impact if done at this stage.

The migration to config.yaml will not be any more complex, mapping to the new `content_type` and `endpoint` fields is trivial.

I think we should take this opportunity to simplify.

jack-berg:
Resolves #103.

- Split `otlp` exporter into `otlp_http` for HTTP transport, and `otlp_grpc` for gRPC transport
- Remove `otlp.protocol` property
- Remove `otlp.endpoint` property from required, since there is now a reasonable default. 
- `otlp_http` and `otlp_grpc` can both be set to null. Since there are no longer any required properties, its reasonable for a user to specify `otlp_http:`, `otlp_grpc:`. 
- The `.insecure` property is only applicable to `otlp_grpc`
- Add new `encoding` property with values `proto` or `json` for `otlp_http`. Default is `proto`. 

Unrelated changes bundled in:
- Rename titles of exporter types to include `*Exporter` prefix
  - Console -> ConsoleExporter
  - Zipkin -> ZipkinExporter
  - Prometheus -> PrometheusExporter

I've prototyped this change in [opentelemetry-java here](https://github.com/jack-berg/opentelemetry-java/commit/9d0c25ea9cc769f7f95b60f08e1504d7a3a4af70).

### DIFF
--- a/schema/common.json
+++ b/schema/common.json
@@ -36,9 +36,10 @@
                 "name", "value"
             ]
         },
-        "BaseOtlpHttpExporter": {
-            "title": "BaseOtlpHttpExporter",
-            "type": "object",
+        "OtlpHttpExporter": {
+            "title": "OtlpHttpExporter",
+            "type": ["object", "null"],
+            "additionalProperties": false,
             "properties": {
                 "endpoint": {
                     "type": ["string", "null"]
@@ -77,9 +78,10 @@
                 }
             }
         },
-        "BaseOtlpGrpcExporter": {
-            "title": "BaseOtlpGrpcExporter",
-            "type": "object",
+        "OtlpGrpcExporter": {
+            "title": "OtlpGrpcExporter",
+            "type": ["object", "null"],
+            "additionalProperties": false,
             "properties": {
                 "endpoint": {
                     "type": ["string", "null"]
@@ -114,8 +116,9 @@
                 }
             }
         },
-        "BaseConsoleExporter": {
-            "type": ["object", "null"]
+        "ConsoleExporter": {
+            "type": ["object", "null"],
+            "additionalProperties": false
         }
     }
 }
--- a/schema/common.json
+++ b/schema/common.json
@@ -36,16 +36,53 @@
                 "name", "value"
             ]
         },
-        "Otlp": {
+        "BaseOtlpHttpExporter": {
+            "title": "BaseOtlpHttpExporter",
             "type": "object",
-            "additionalProperties": false,
             "properties": {
-                "protocol": {
-                    "type": "string",
-                    "pattern": "^(http|grpc)\\/(protobuf|json)"
-                },
                 "endpoint": {
-                    "type": "string"
+                    "type": ["string", "null"]
+                },
+                "certificate": {
+                    "type": ["string", "null"]
+                },
+                "client_key": {
+                    "type": ["string", "null"]
+                },
+                "client_certificate": {
+                    "type": ["string", "null"]
+                },
+                "headers": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "common.json#/$defs/NameStringValuePair"
+                    }
+                },
+                "headers_list": {
+                    "type": ["string", "null"]
+                },
+                "compression": {
+                    "type": ["string", "null"]
+                },
+                "timeout": {
+                    "type": ["integer", "null"],
+                    "minimum": 0
+                },
+                "encoding": {
+                    "type": ["string", "null"],
+                    "enum": [
+                        "proto",
+                        "json"
+                    ]
+                }
+            }
+        },
+        "BaseOtlpGrpcExporter": {
+            "title": "BaseOtlpGrpcExporter",
+            "type": "object",
+            "properties": {
+                "endpoint": {
+                    "type": ["string", "null"]
                 },
                 "certificate": {
                     "type": ["string", "null"]
@@ -75,16 +112,10 @@
                 "insecure": {
                     "type": ["boolean", "null"]
                 }
-            },
-            "required": [
-                "endpoint",
-                "protocol"
-            ],
-            "title": "Otlp"
+            }
         },
-        "Console": {
-            "type": ["object", "null"],
-            "additionalProperties": false
+        "BaseConsoleExporter": {
+            "type": ["object", "null"]
         }
     }
 }
--- a/schema/common.json
+++ b/schema/common.json
@@ -72,7 +72,7 @@
                 "encoding": {
                     "type": ["string", "null"],
                     "enum": [
-                        "proto",
+                        "protobuf",
                         "json"
                     ]
                 }
--- a/schema/logger_provider.json
+++ b/schema/logger_provider.json
@@ -66,13 +66,13 @@
             "maxProperties": 1,
             "properties": {
                 "otlp_http": {
-                    "$ref": "#/$defs/OtlpHttpLogRecordExporter"
+                    "$ref": "common.json#/$defs/OtlpHttpExporter"
                 },
                 "otlp_grpc": {
-                    "$ref": "#/$defs/OtlpGrpcLogRecordExporter"
+                    "$ref": "common.json#/$defs/OtlpGrpcExporter"
                 },
                 "console": {
-                    "$ref": "#/$defs/ConsoleLogRecordExporter"
+                    "$ref": "common.json#/$defs/ConsoleExporter"
                 }
             },
             "patternProperties": {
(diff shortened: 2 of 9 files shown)


schema/logger_provider.json
```diff
@@ -65,11 +65,14 @@
             "minProperties": 1,
             "maxProperties": 1,
             "properties": {
-                "otlp": {
-                    "$ref": "common.json#/$defs/Otlp"
+                "otlp_http": {
+                    "$ref": "common.json#/$defs/OtlpHttpExporter"
+                },
+                "otlp_grpc": {
```
jack-berg, line 71: There's a question of whether we should align with the names of exporters in the collector, which follows the same strategy of splitting the exporter by transport:
- [otlp](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlpexporter) for gRPC
- [otlphttp](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlphttpexporter) for HTTP (proto and json)

I've chosen to _not_ align with the collector for a couple of reasons:
1. The collector follows a different case naming convention than this project. We [prefer lower snake case](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#property-name-case), while the collector prefers something that I've seen called [flatcase or memblecase](https://stackoverflow.com/questions/17326185/what-are-the-different-kinds-of-cases). We'd have to break out case policy to align with the collector.
2. The collector prefers gRPC, which is an out of date default. By omitting a suffix on the grpc variant (i.e. `otlp` vs `otlpgrpc` and `otlphttp` instead of `otlp`), the collector is expressing a subtle preference for grpc. This was the original default for OTLP but OTEL has since switched course preferring HTTP. In the naming convention I've chosen, HTTP and gRPC are presented as equals. I don't think we should express a preference, but if we did, it should be for HTTP (i.e. `otlp` for HTTP and `otlpgrpc` for gRPC). This would be confusing for collector users, so better to explicitly break rank.

schema/common.json
```diff
@@ -36,16 +36,55 @@
...
                 },
+                "timeout": {
+                    "type": ["integer", "null"],
+                    "minimum": 0
+                },
+                "encoding": {
+                    "type": ["string", "null"],
+                    "enum": [
+                        "proto",
+                        "json"
```
marcalff, line 76: Given that the existing options are `http/protobuf` and `http/json`, I think we should spell out `protobuf` instead of just `proto`, for the encoding enum.

schema/meter_provider.json
```diff
@@ -196,30 +214,64 @@
...
                     "minimum": 0
                 },
-                "temporality_preference": {
+                "encoding": {
                     "type": ["string", "null"],
                     "enum": [
-                        "cumulative",
-                        "delta",
-                        "low_memory"
+                        "proto",
```
marcalff, line 220: Likewise, `protobuf` instead of `proto`.

marcalff (commented):
LGTM, with only a small comment on naming (proto -> protobuf)

marcalff (approved):
LGTM.

codeboten (commented):
I know we're still not providing backwards compatibility guarantees, but changes like this (and previously the headers change) should be captured somewhere as being breaking changes.

Maybe it's easy enough to label the title of this PR `breaking: refactor OTLP exporter` so that at release time these are listed in the release notes

jack-berg:
Wondering if any other @open-telemetry/configuration-approvers care about this / plan on reviewing it. Happy to leave it open longer if more time is needed.

brettmc:
> care about this / plan on reviewing it. Happy to leave it open longer if more time is needed

I'd like a day or two to review this please

jack-berg:
At this point, might as well leave it open until the new year. No rush as this doesn't hold up anything between now and then.

brettmc:
Here's an alternative for consideration. I also included the in-development [otlp file exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/file-exporter.md)

<details>
<summary>
config.yaml
</summary>

```yaml
tracer_provider:
    processors:
        - batch:
            exporter:
                otlp:
                    transport:
                        http:
                            encoding: protobuf
                            headers:
                                - name: api-key
                                  value: "1234"
                            endpoint: http://localhost:4318/v1/traces
                            timeout: 10000
        - batch:
            exporter:
                otlp:
                    transport:
                        http:
                            encoding: json
                            endpoint: http://localhost:4318/v1/traces
                            compression: gzip
                            timeout: 10000
        - batch:
            exporter:
                otlp:
                    transport:
                        grpc:
                            insecure: true
                            endpoint: localhost:4317
                            certificate: /app/cert.pem
                            client_certificate: /app/cert.pem
        - batch:
            exporter:
                otlp:
                    transport:
                        file:
                            stream: stdout #default value
        - batch:
            exporter:
                otlp:
                    transport:
                        file:
                            stream: /path/to/traces.jsonl
```

</details>

The `transport` key is possibly not required, but I think it enhances readability...

jack-berg:
> Here's an alternative for consideration. I also included the in-development [otlp file exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/file-exporter.md)

A quick summary of this proposal: All exporters which use OTLP message encodings are considered the same exporter from a schema perspective, `otlp`, and the differences between them are shifted to the lower level in `otlp.transport`. Results in an additional level that every user will encounter. 

This PR in its current form goes the opposite direction, with distinct OTLP exporters for each transport, with names following the form `otlp_{transport}`, where `{transport}` is one of `file`, `http`, or `grpc`. 

Since your proposal and this PR don't differ in the set of transports or in the properties for each, its the difference between:

```
otlp_http: ..
otlp_grpc: ...
otlp_file: ...
```

and 

```
otlp:
  transport:
    http:
otlp:
  transport:
    grpc:
otlp:
  transport:
    file:
```

Its appears to be a question of whether the delimiter between "otlp" and `{transport}` should be `_` (i.e. this PR) or `\ntransport:\n` (your proposal).

Personally, I prefer the `_` delimiter. Reasons:

- Reduced nesting improves the UX (in my opinion). 
- Mirrors the collector's organization, and to probably SDK organization - how many SDKs treat the OTLP file exporter as a configuration option for a single OTLP exporter vs. a separate concept? For example, opentelemetry-java has different exporters for each transport.
- The proposal suggests that there should be other properties that are siblings to `transport`, but its hard to imagine what those would be since all properties will likely be dependent on the transport.

brettmc (approved):

jack-berg:
@codeboten any last words before I merge?

MrAlias (approved):
:medal_sports:

PR #147, opened 2024-12-05
Add guidance around use of polymorphic types

jack-berg:
On multiple occasions I've tried to use `allOf` to model polymporphic types only to be reminded that the jsonschema2pojo tool we're using in opentelemetry-java [doesn't support anyOf, allOf, oneOf](https://github.com/joelittlejohn/jsonschema2pojo/issues/1649). I suspect this is the case in other languages as well.

I propose we provide modeling guidance to explicitly avoid this tempting trap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,70 @@ If a property is _not_ required, it should include a [comment](./CONTRIBUTING.md
 
 If a property `type` includes `null`, it must include a [comment](./CONTRIBUTING.md#description-generation) describing the semantics when the value is `null`. It's common for properties with primitive types to allow `null`. `object` types allow `null` if no properties are required and the presence of the property key is meaningful. 
 
+### Polymorphic types
+
+JSON schema's [schema composition](https://json-schema.org/understanding-json-schema/reference/combining) keywords (`allOf`, `anyOf`, `oneOf`) offer a tempting mechanism for object-oriented style inheritance and polymorphic patterns. However, JSON schema code generation tools may struggle or not support these keywords. Therefore, these keywords should be used judiciously, and should not be used to extend `object` types.
+
+For example:
+
+```json
+{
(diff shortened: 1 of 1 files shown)


marcalff:
> On multiple occasions I've tried to use `allOf` to model polymporphic types only to be reminded that the jsonschema2pojo tool we're using in opentelemetry-java [doesn't support anyOf, allOf, oneOf](https://github.com/joelittlejohn/jsonschema2pojo/issues/1649). I suspect this is the case in other languages as well.
> 
> I propose we provide modeling guidance to explicitly avoid this tempting trap.

I Agree.

In case of C++, the available tooling is so poor (or nonexistent) that:

* we are not even using a schema validator
* we are not using code generated from a schema

Validation of yaml nodes against the schema is done ad-hoc.

marcalff (approved):
LGTM.

Not related to this PR, but I think it is time to split README.md into multiple documents, and have a docs/ subdirectory.

carlosalberto (approved):

jack-berg:
> Validation of yaml nodes against the schema is done ad-hoc.

That's too bad on the tooling. For what its worth, once the schema is stable, I'm strongly considering ripping out the code generation tooling from `opentelemetry-java` and maintaining the (previously) generated code by hand instead. Is very similar to the problem with exposing generated proto bindings from `opentelemetry-proto` as part of your public stable API: you can't guarantee that the code generation tooling won't change in a way that violates the guarantees you make for stable APIs.

codeboten (approved):

codeboten:
Re. tooling, on the go side i ended up having to split the model generation from the parsing logic, as the parsing code needed to be updated in a couple of places.

I guess this is a good opportunity to help move the json schema ecosystem forward where possible.

PR #148, opened 2024-12-05
Fix MetricReader invalid configurations

### LINKED ISSUE #138: MetricReader allows invalid configurations
The following schema:

```
        "MetricReader": {
            "type": ["object", "null"],
            "minProperties": 1,
            "maxProperties": 2,
            "properties": {
                "periodic": {
                    "$ref": "#/$defs/PeriodicMetricReader"
                },
                "pull": {
                    "$ref": "#/$defs/PullMetricReader"
                },
                "producers": {
                    "type": "array",
                    "items": {
                        "$ref": "#/$defs/MetricProducer"
                    }
                }
            }
        },
```

is causing semantic issues.

Previously, it contained only "periodic" and "pull" properties, with max properties = 1, meaning a metric reader is either periodic or pull.

When producers was added, max properties was changed to 2.
This allows "periodic" + "producers", or "pull" + "producers", which is the intent (as I understand).

This however also allows to have both "periodic" and "pull", which are no longer mutually exclusive.

This opens the schema validation to unintended configurations, pushing this downstream to the SDK implementation.

If this combination is not allowed, this must be clarified explicitly:

* for end users, to know what to expect
* for SDK implementors, to test explicitly this case, so it can be handled correctly.

Found while implementing config.yaml for C++:

* https://github.com/open-telemetry/opentelemetry-cpp/pull/2518

marcalff:
Fixes #138

* Move `producers` from `MetricReader` to `PeriodicMetricReader`
* Move `producers` from `MetricReader` to `PullMetricReader`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Improved file_format documentation. [#137](https://github.com/open-telemetry/opentelemetry-configuration/pull/137)
 * Periodic exporter interval default value is inconsistent [#143](https://github.com/open-telemetry/opentelemetry-configuration/pull/143)
+* Fix MetricReader invalid configurations [#148](https://github.com/open-telemetry/opentelemetry-configuration/pull/148)
 
 ## [v0.3.0] - 2024-05-08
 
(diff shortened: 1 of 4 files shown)


jack-berg (approved):
This is a better way to model this. There's little / no advantage to having the `producers` a property of `MetricReader` vs. `PeriodicMetricReader` / `PullMetricReader`. 

Thanks!

PR #149, opened 2024-12-09
Add release step to update file_format in examples

### LINKED ISSUE #132: Ensure we don't forget to update file_format in examples in future releases
Let's add process to ensure we don't have another instance of #129.

Could be as simple as updating the [release process](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/RELEASING.md), or maybe we add some sort of automation to ensure we don't forget.

jack-berg:
Resolves #132.


codeboten (approved):

marcalff (approved):
LGTM.

PR #151, opened 2024-12-09
Change various usages of minimum to exclusiveMinimum

jack-berg:
Related to https://github.com/open-telemetry/opentelemetry-specification/issues/4283, but not blocked by the resolution over there. 

With #142, the main principle of what constitutes an allowed change is whether it makes validation more or less restrictive. In reviewing our use of the JSON schema [minimum keyword](https://json-schema.org/understanding-json-schema/reference/numeric#range), which has inclusive semantics, I found a variety of areas which may come back to bite us. In all these cases, the minimum is currently set to 0. But if the user sets the property value to zero, the semantics are unclear.

Best to prohibit zero in these cases, since we can later loosen the restriction to allow zero but won't be able to make it more strict.

Cases:

- Processor (log, span) batch size
- Processor (log, span) queue size
~~Processor (log, span, periodic metric reader) exporter timeout~~
~~Processor (log, span, periodic metric reader) export internal~~
~~Exporter (otlp, zipkin) timeout~~
~~Jaeger remote sampler fetch interval~~

**UPDATE (1/13/2025)**: Updated to reflect the state of things in [Spec PR#4331](https://github.com/open-telemetry/opentelemetry-specification/pull/4331) which now seems poised to merge. Differences from what I originally proposed:

* Durations can be zero.
* Timeouts can be zero, with zero indicating indefinite.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -14,10 +14,10 @@ file_format: "0.3"
 disabled: false
 # Configure general attribute limits. See also tracer_provider.limits, logger_provider.limits.
 attribute_limits:
-  # Configure max attribute value size.
+  # Configure max attribute value size. Value must be non-negative.
   # If omitted or null, there is no limit.
   attribute_value_length_limit: 4096
-  # Configure max attribute count.
+  # Configure max attribute count. Value must be non-negative.
   # If omitted or null, 128 is used.
   attribute_count_limit: 128
 # Configure logger provider.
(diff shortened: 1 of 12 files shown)


examples/kitchen-sink.yaml
```diff
@@ -27,16 +27,16 @@ logger_provider:
   processors:
     - # Configure a batch log record processor.
       batch:
-        # Configure delay interval (in milliseconds) between two consecutive exports.
+        # Configure delay interval (in milliseconds) between two consecutive exports. Value must be > 0.
         # If omitted or null, 1000 is used.
         schedule_delay: 5000
```
jack-berg, line 32: Arguably a schedule delay of 0 is valid. Made this change anyway to prompt debate.

marcalff (approved):
LGTM

examples/kitchen-sink.yaml
trask: I agree with your comment about 0 being valid https://github.com/open-telemetry/opentelemetry-specification/pull/4331#discussion_r1878484937

jack-berg:
Status update: waiting to see how https://github.com/open-telemetry/opentelemetry-specification/pull/4331 shakes out. Based on this [conversation](https://github.com/open-telemetry/opentelemetry-specification/pull/4331#discussion_r1881034983), it appears the preference is to interpret a timeout value of zero to mean indefinite. Although not idea if given a fresh start, its a reasonable conclusion to reach given diverging implementations.

We have somewhat of a fresh start in declarative config, but are tied back to the env vars config schema via env var substitution.

jack-berg:
**UPDATE (1/13/2025)**: Updated to reflect the state of things in [Spec PR#4331](https://github.com/open-telemetry/opentelemetry-specification/pull/4331) which now seems poised to merge. Differences from what I originally proposed:

* Durations can be zero.
* Timeouts can be zero, with zero indicating indefinite.

Given these changes, I've cleared the approvals. Please take a look @open-telemetry/configuration-maintainers.

brettmc (approved):

marcalff (approved):
LGTM, my only comment is editorial.

Lines like:
```
        # Configure maximum allowed time (in milliseconds) to export data. Value must be non-negative. A value of 0 indicates indefinite.
```

are very long in the yaml file itself.

Could type_description.yaml break this into separate lines, by adding an empty line as delimiter ?

The default configuration in examples will be copied and pasted a lot when people will use this, so a reasonable max size for a line will make it easier to consume.

It would be good to

jack-berg:
> Could type_description.yaml break this into separate lines, by adding an empty line as delimiter ?

👍 yes I can do this

PR #155, opened 2025-01-14
Object and enum types should be defined in $defs

jack-berg:
More cleanup as we approach stability.

Here I ensure the schema consistently uses JSON schema's schema / subschema options with a new schema modeling rule:

> In order to promote stylistic consistency and allow for reuse of concepts, `object` and `enum` types should be defined in either as a top level schema document or as a subschema in a schema document's `$defs`.

The stylistic consistency part is obvious.

"allow for reuse of concepts" is less obvious. When a type is defined in `$defs` we can reference as many times as needed in properties. When a type is defined directly in a property (or inline if you prefer), that type can only describe the requirements for that single property.

This guidance applies to  `object` and `enum` types. This means that excluded are all primitives and arrays.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,18 +172,74 @@ Another example:
         ]
       },
       "type": {
-        "type": ["string", "null"],
-        "enum": [null, "string", "bool", "int", "double", "string_array", "bool_array", "int_array", "double_array"]
+        "$ref": "#/$defs/AttributeType"
       }
     },
     "required": [
       "name", "value"
+    ]
+  },
+  "AttributeType": {
+    "type": ["string", "null"],
+    "enum": [
+      null,
+      "string",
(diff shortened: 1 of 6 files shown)


brettmc (approved):

README.md
```diff
@@ -172,19 +172,75 @@ Another example:
...
+}
+```
+
+Or a property can reference a subschema residing in a schema document's [$defs](https://json-schema.org/understanding-json-schema/structuring#defs):
+
+```json
+{
+  "properties": {
+    "shape": {
+      "$ref": "#/$defs/Sampler"
```
marcalff, line 227: Sampler -> Shape

marcalff (approved):
LGTM, see minor changes.

PR #157, opened 2025-01-14
Add guidance around use of title and description keywords

jack-berg:
Related to #142, which references that "Type [title](https://json-schema.org/understanding-json-schema/reference/annotations) will not change."

I still agree with this, but don't think we should define type titles using the `title` keyword. Instead, the `title` should be sourced from the key values withing `$defs`. For example, given the following from `meter_provider.json`, the periodic metric reader's title is `PeriodicMetricReader`. See PR content for reasoning.

```
   ...
    "$defs": {
        "PeriodicMetricReader": {
            "type": "object",
            "properties": {
                "interval": {
                    "type": ["integer", "null"],
                    "minimum": 0
                },
                "timeout": {
                    "type": ["integer", "null"],
                    "minimum": 0
                },
                "exporter": {
                    "$ref": "#/$defs/PushMetricExporter"
                },
                "producers": {
                    "type": "array",
                    "items": {
                        "$ref": "#/$defs/MetricProducer"
                    }
                }
            },
            "required": [
                "exporter"
            ],
            "title": "PeriodicMetricReader"
        },
   ...
```

### DIFF
--- a/schema/common.json
+++ b/schema/common.json
@@ -1,7 +1,6 @@
 {
     "$id": "https://opentelemetry.io/otelconfig/common.json",
     "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "title": "Common",
     "$defs": {
         "IncludeExclude": {
             "type": "object",
@@ -37,7 +36,6 @@
             ]
         },
         "OtlpHttpExporter": {
-            "title": "OtlpHttpExporter",
             "type": ["object", "null"],
             "additionalProperties": false,
             "properties": {
@@ -79,7 +77,6 @@
             }
         },
         "OtlpGrpcExporter": {
-            "title": "OtlpGrpcExporter",
             "type": ["object", "null"],
             "additionalProperties": false,
             "properties": {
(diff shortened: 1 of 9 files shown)


schema/common.json
```diff
@@ -1,7 +1,6 @@
 {
     "$id": "https://opentelemetry.io/otelconfig/common.json",
     "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "title": "Common",
```
jack-berg, line 4: Notably, removing `title` in top level schema like this, which collect types for re-use but which aren't referenced anywhere in the schema directly tidies up the result of code generation. 

Previously, the `opentelemetry-java` implementation generated a `CommonModel` type. It wasn't used anywhere, but it still was generated. With this change, its goes away as it should. I could probably find some way to explicitly omit generation for these kind of types, but it is a nice side affect of what seems like a sensible change for other reasons.

brettmc (approved):

marcalff (approved):
LGTM.

PR #165, opened 2025-01-24
fixing broken link to env var substitution spec

### LINKED ISSUE #163: is environment variable substitution with `env:` prefix supported?
https://github.com/open-telemetry/opentelemetry-configuration/blob/main/validator/README.md#usage mentions: 

> Environment variable substitution is supported with the syntax ${VARIABLE} or ${env:VARIABLE}.

None of the examples do this, and there's a broken link to the spec which might have given more detail. What we do have looks like it arrived in https://github.com/open-telemetry/opentelemetry-specification/pull/3974

brettmc:
fix link to the spec data model, which defines the requirements for env var substitution

Closes #163


codeboten (approved):
Thanks!

PR #173, opened 2025-02-18
feat: add minimum length requirement to propagator configuration

### LINKED ISSUE #172: propagator name should not be a blank string
If a user sets the following configuration, I would expect a configuration error to be returned:

```
propagator:
  composite: [""]
```

This would ensure that implementations follow the same pattern. The configuration above is currently valid as defined by the schema https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/propagator.json, I suggest we add a NonEmptyString type and use that as the type for propagator configuration.

codeboten:
BREAKING CHANGE: This is potentially breaking any implementation that ignores a blank value for a propagator today.

Fixes #172

### DIFF
--- a/schema/common.json
+++ b/schema/common.json
@@ -128,6 +128,10 @@
         "ConsoleExporter": {
             "type": ["object", "null"],
             "additionalProperties": false
+        },
+        "NonEmptyString": {
+            "type": "string",
+            "minLength": 1
         }
     }
 }
--- a/schema/propagator.json
+++ b/schema/propagator.json
@@ -6,7 +6,7 @@
         "composite": {
             "type": "array",
             "items": {
-                "type": ["string"]
+                "$ref": "common.json#/$defs/NonEmptyString"
             }
         }
     },


marcalff:
Maybe the issue is that propagators should not be an array of strings in the first place.

Instead of:
```
propagator:
  composite: [ tracecontext, baggage, b3, b3multi, jaeger, xray, ottrace ]
```

Can we have:
```
propagator:
  - tracecontext
  - baggage
  - b3
  - b3multi
  - jaeger
  - xray
  - ottrace
```

For every extension point (exporters, processors, etc), the name appears as a yaml node.

For propagators, the name appears as data, this does not follows the general pattern.

Custom propagators are extension points in my understanding, because they will need to be registered in the SDK, they are not just "strings".

jack-berg:
@marcalff I think you mean:

```
propagator:
  - tracecontext:
  - baggage:
  - b3:
  - b3multi:
  - jaeger:
  - xray:
  - ottrace:
```

jack-berg:
> For propagators, the name appears as data, this does not follows the general pattern.

I agree we should fix this. 

Should also probably add backwards compatibility with `OTEL_PROPAGATORS` following our [data modeling and environment variable substitution](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#data-modeling-and-environment-variable-substitution) rule.

End config might look like:

```
propagator:
  composite:
      - tracecontext:
      - baggage:
      - b3:
      - b3multi:
      - jaeger:
      - xray:
      - ottrace:
  # The value is a list of comma separated key-value pairs matching the format of OTEL_PROPAGATORS. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration for details.
  # Merged with .propagator.composite. If an entry appears in both, .propagator.composite takes priority.
  composite_list: ${OTEL_PROPAGATORS}
```

marcalff (changes requested):
Blocking this fix in favor of PR #187
Having dedicated yaml nodes for extension points is more aligned with the existing codebase.

codeboten:
Closing as addressed in https://github.com/open-telemetry/opentelemetry-configuration/pull/187

PR #187, opened 2025-03-06
Refactor propagator schema

jack-berg:
See [comment](https://github.com/open-telemetry/opentelemetry-configuration/pull/173#issuecomment-2694969495).

Supersedes #173.

This refactors `.propagator.composite` from being an array of string entries to an array of objects. Each object has a single key identifying the name of the propagator.

This aligns propagator with the data modeling of other SDK extension points (processors, exporters, readers, sampler) where the extension point identifier is a key, and the object's value is passed as a node to configure that extension point. In the case of propagators, none of the built-in propagators currently accept configuration, but its conceivable that they may in the future.

The propagator comments have also been changed to indicate that omitting `.propagator` or having an empty `.propagator.composite` list results in a noop propgator. This aligns propagator config with the "what you see is what you get" philosophy used elsewhere in declarative config data modeling.

Finally, a new `.propagator.composite_list` property has been added with the same semantics as `OTEL_PROPAGATORS` env var. The comments indicate a merge semantic where entries from `.propagator.composite_list` are appended to `.propagator.composite` (with duplicates removed) when both are present. The ordering here is important because order that propagators are registered is signficiant, particularly when extracting context.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -497,11 +497,29 @@ meter_provider:
   # If omitted or null, trace_based is used.
   exemplar_filter: trace_based
 # Configure text map context propagators.
-# If omitted, tracecontext and baggage are used.
+# If omitted, a noop propagator is used.
 propagator:
-  # Configure the set of propagators to include in the composite text map propagator. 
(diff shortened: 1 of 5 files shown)


marcalff (approved):
LGTM, thanks for the fix.

PR #188, opened 2025-03-10
Refactor resource detection to allow config of named resource detectors

### LINKED ISSUE #119: make it possible to configure named resource detectors
It would be great to have the capability to configure resource detectors via the configuration file.

For example, the Collector needs to be able to specify that the environment variable resource detector should be used to address https://github.com/open-telemetry/opentelemetry-collector/issues/10909

jack-berg:
Resolves #119.

Resource detection is currently naive:

- We only have the ability to specify an allow list / disallow list of attributes that come from resource detectors.
- Its assumed / implied that all resource detectors are enabled, which conflicts with declarative config's "what you see is what you get" philosophy.
- There is no ability to pass configuration to a particular resource detector.
- There is no ability to influence the order of resource detectors, so if two provide the same attribute, you're at the mercy of whatever order they were applied in.

This PR refactors resource detection to add more expressiveness and better accommodate a wide range of use cases.

Let's take a look at an example config after this refactor:
```
resource:
  attributes:
    - name: service.name
       value: my-service
  detection/development:
    detectors:
      - container:
      - os:
      - host:
      - process:
    attributes:
      excluded:
        - process.command_line
```

Most users will want to set a static set of resource attributes, especially `service.*`, but resource detection is an important part of the otel ecosystem. You opt into resource detection by setting `resource.detection/development` and configuring to suit. 

`.detectors` is an array of objects, each with a single key referring to the name of a detector, and an object value passed to that detector to optionally configure it. This aligns resource detectors which the design of other SDK extension components.

The ability to include / exclude attributes provided by the collection of all resource detectors is still available.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -747,13 +747,6 @@ resource:
   # This type is in development and subject to breaking changes in minor versions.
   # If omitted or null, resource detection is disabled.
   detection/development:
-    # Configure the strategy for enabling resource detectors.
-    # Values include:
-    #  * opt_in: Resource detectors configured in .detectors are enabled and applied in the order specified. Other resource detectors are disabled. 
-    #  * enable_all: All resource detectors are enabled and applied in an unspecified order. Configuration in .detectors is passed to respective detectors.
-    #  * disable_all: All resource detectors are disabled, including any configured in .detectors.
-    # If omitted or null, opt_in is used.
-    enabled_detectors_strategy: opt_in
     # Configure attributes provided by resource detectors.
     attributes:
       # Configure list of attribute key patterns to include from resource detectors.
@@ -770,9 +763,9 @@ resource:
       # If omitted, .included attributes are included.
       excluded:
         - process.command_args
-    # Configure resource detectors. See also enabled_detectors_strategy.
+    # Configure resource detectors.
     # Resource detector names are dependent on the SDK language ecosystem. Please consult documentation for each respective language. 
-    # If omitted or null and enabled_detectors_strategy=opt_in, no resource detectors are enabled.
+    # If omitted or null, no resource detectors are enabled.
     detectors:
       - # Enable the container resource detector.
         # Note, the key "container" is an example and detector names may vary by SDK language ecosystem.
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -743,9 +743,17 @@ resource:
   # The value is a list of comma separated key-value pairs matching the format of OTEL_RESOURCE_ATTRIBUTES. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration for details.
   # If omitted or null, no resource attributes are added.
   attributes_list: "service.namespace=my-namespace,service.version=1.0.0"
-  # Configure resource detectors.
+  # Configure resource detection.
   # This type is in development and subject to breaking changes in minor versions.
-  detectors/development:
+  # If omitted or null, resource detection is disabled.
+  detection/development:
+    # Configure the strategy for enabling resource detectors.
+    # Values include:
+    #  * opt_in: Resource detectors configured in .detectors are enabled and applied in the order specified. Other resource detectors are disabled. 
+    #  * enable_all: All resource detectors are enabled and applied in an unspecified order. Configuration in .detectors is passed to respective detectors.
+    #  * disable_all: All resource detectors are disabled, including any configured in .detectors.
+    # If omitted or null, opt_in is used.
+    enabled_detectors_strategy: opt_in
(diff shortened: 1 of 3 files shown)


examples/kitchen-sink.yaml
```diff
@@ -762,6 +770,22 @@ resource:
...
+        container:
+      - # Enable the host resource detector.
+        # Note, the key "host" is an example and detector names may vary by SDK language ecosystem.
+        host:
+      - # Enable the os resource detector.
+        # Note, the key "os" is an example and detector names may vary by SDK language ecosystem.
+        os:
+      - # Enable the process resource detector.
+        # Note, the key "process" is an example and detector names may vary by SDK language ecosystem.
+        process:
```
jack-berg, line 788: The lack of a well known list of standard resource detector names is the biggest problem with this design. I'd like to fix that by adding such a list to the [resource specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment), along with general guidance that resource detectors be named according to the root namespace of attributes they typically provide. For example, a resource detector that provides `os.*` attributes should be named `os`, one that provides `process.*` should be named `process`, etc.

examples/kitchen-sink.yaml
```diff
@@ -743,9 +743,17 @@ resource:
...
-  detectors/development:
+  # If omitted or null, resource detection is disabled.
+  detection/development:
+    # Configure the strategy for enabling resource detectors.
+    # Values include:
+    #  * opt_in: Resource detectors configured in .detectors are enabled and applied in the order specified. Other resource detectors are disabled. 
+    #  * enable_all: All resource detectors are enabled and applied in an unspecified order. Configuration in .detectors is passed to respective detectors.
+    #  * disable_all: All resource detectors are disabled, including any configured in .detectors.
+    # If omitted or null, opt_in is used.
+    enabled_detectors_strategy: opt_in
```
jack-berg, line 756: Initially, I thought that resource detectors should be purely opt-in. But the convenience of being able to enable all that are available without needing to know what they are named is too good of a user experience to ignore, so I added `enable_all` accordingly. `disable_all` is a bit of syntactic sugar, since you can achieve the same effect by setting `opt_in` and `detectors: []`, but its nice to be able to use an environment variable to turn off all resource detectors. I.e. given the following:

```
resource:
  detection:
    enabled_detectors_strategy: ${OTEL_ENABLED_RESOURCE_DETECTORS_STRATEGY:-opt_in}
    detectors:
      - containers:
      - process:
```

You can set `OTEL_ENABLED_RESOURCE_DETECTORS_STRATEGY=disable_all` to disable detection and override `.detectors`, without changing the content of your config file.

jack-berg:
@trask curious what you think of this proposal. It adds to the ideas in the comments in #119, but enables the common use case of wanting to enable all resource detectors, and with a syntax that I think will be fairly intuitive.

examples/kitchen-sink.yaml
marcalff: > For example, a resource detector that provides `os.*` attributes should be named `os`

I would turn it differently:

A resource detector named `os` should provide `os.*` attributes.

This is ok because the name, `os`, is a prefix for attributes.

It should be also possible to have a resource detector named `os_simulated`, `qa`, or `my_fake_detector`, that also provides `os.*` attributes.

examples/kitchen-sink.yaml
marcalff: I have some reserves and concerns about the `enabled_detectors_strategy: enable_all` option.

This enables all ... that is registered in the detector extension registry.

Two applications running with the same config.yaml may behave differently, because the content of the registry (the list of detectors known to the SDK) differ, even for the same SDK implementation, due to what was/was not registered.

I fear that this is too unpredictable, hard to explain, and hard to troubleshoot, and will cause more issues that it solves.

Listing the detectors to use should not be that hard, and makes it predictable, discoverable, and explicit.

Assuming `enable_all` is dropped, this:

```
resource:
  detection:
    enabled_detectors_strategy: ${OTEL_ENABLED_RESOURCE_DETECTORS_STRATEGY:-opt_in}
```

can be simplified to:

```
resource:
  detection:
    # true or false
    enabled: ${OTEL_RESOURCE_DETECTION_ENABLED:-true}
```

I like the idea of a kill switch to turn off detection, while keeping the config in the yaml file.

jack-berg: I ended up pulling the `enabled_detectors_strategy` property for now, given that we can add it later.

I also omitted the suggested `.resource.detection.enabled` because a user can effectively turn off resource detection via env vars by setting an entry in `.resource.detection.attributes.excluded` to `*`, i.e.:

```
resource:
  detection:
    attributes:
      excluded:
        - *
```

Can always go back and either of these approaches later, but let's start with the simplest thing and revisit if needed.

marcalff (approved):
LGTM.

PR #197, opened 2025-04-07
docs: document implementation matrix

codeboten:
The idea w/ this matrix is to provide end users a clear picture of what is implemented for each language implementation. Ideally, the matrix will be:

1. auto generated from the kitchen-sink
2. automatically filled in for every implementation via a test

In the short term, the automation for fill in the implementation details could be manual, but ideally that would be short-lived for the sake of users and implementors.

### DIFF
--- a/matrix.md
+++ b/matrix.md
@@ -1,0 +1,45 @@
+# Implementation matrix
+
+The following table shows the implementation of each feature supported by the
+configuration model:
+
+
+| Top Level | C++|Java|Go|PHP |
+| --------- | -|-|-|- | 
+| `file_format` | -|-|-|- | 
+| `disabled` | -|-|-|- | 
+| `log_level` | -|-|-|- | 
+| `attribute_limits` | -|-|-|- | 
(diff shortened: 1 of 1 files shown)


marcalff:
@jack-berg Could you check your demo branch ? I can't see the script or the generated markdown. Thanks.

jack-berg:
Oops looks like I had some copy paste errors on the links. Fixed. 

This still needs a decent amount of work before its ready for PR, but I think its conceptually right.

marcalff:
> Oops looks like I had some copy paste errors on the links. Fixed.
> 
> This still needs a decent amount of work before its ready for PR, but I think its conceptually right.

Thanks, I can see it now.

Some suggestion below

Instead of presenting:

* a table with "Property, Type, Description, C++, C#, erlang, etc"

consider presenting:

* a table with "Property, Type, Description"
* a table with "Property, C++, C#, erlang, etc"

Also, extension points are missing.

In Samplers for example, I expect an entry for extension points, and the possibility for each SIG to document if extension points are suported or not.

PR #215, opened 2025-06-16
Set allow-dirty for cargo dist

### LINKED ISSUE #214: Rennovate pinned dependencies break build
Looks like this broke the build ([example](https://github.com/open-telemetry/opentelemetry-configuration/actions/runs/15690656852/job/44204771801?pr=213)) because `.github/workflows/validator-release.yml` is generated by [dist](https://github.com/axodotdev/cargo-dist), and the pinned versions are different than what is generated.

_Originally posted by @jack-berg in https://github.com/open-telemetry/opentelemetry-configuration/issues/204#issuecomment-2978369604_

jack-berg:
Fixes #214.

### DIFF
--- a/.github/workflows/validator-tests.yaml
+++ b/.github/workflows/validator-tests.yaml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Run govulncheck
         run: |
-          make -C validator govulncheck
+#          make -C validator govulncheck
 
       # TODO: Publish releases
       - name: Build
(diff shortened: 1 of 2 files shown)


codeboten (approved):
The govuln failures are addressed in https://github.com/open-telemetry/opentelemetry-configuration/pull/213

.github/workflows/validator-tests.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: Run govulncheck
         run: |
-          make -C validator govulncheck
+#          make -C validator govulncheck
```
jack-berg, line 26: @codeboten disabled so we can merge, and can re-enable in #213

PR #199, opened 2025-04-23
Standardize container, host, process, and service resource detectors

jack-berg:
The concept of named resource detectors with a standard set of known names has been added to the spec in https://github.com/open-telemetry/opentelemetry-specification/pull/4461

PR #220, opened 2025-06-20
Move known resource detectors to properties

jack-berg:
Fixing a modeling error from #199.

Shouldn't have any impact on users.


marcalff (approved):
LGTM

PR #239, opened 2025-07-07
Add `validator` target as dependency for `validator-build-shelltest-image`

MrAlias:
This is a required step for successful completion of the `validator-build-shelltest-image` target.


marcalff (approved):
LGTM

PR #240, opened 2025-07-07
Run docker image as current user

MrAlias:
Do not run as `root`. This ensures the `out` directory is owned by the correct user.


marcalff (approved):
LGTM

PR #244, opened 2025-07-11
add http known methods

zeitlinger:
Relates https://github.com/open-telemetry/opentelemetry-java/issues/7468


zeitlinger:
@jack-berg please have a look

jack-berg (approved):
Looks good but need to wait for corresponding semantic-conventions PR to merge: https://github.com/open-telemetry/semantic-conventions/pull/2504

zeitlinger:
> Looks good but need to wait for corresponding semantic-conventions PR to merge: [open-telemetry/semantic-conventions#2504](https://github.com/open-telemetry/semantic-conventions/pull/2504)

- none of the other config options are currently in semconv
- I think it should be a separate task to move all config options to semconv 
- based on https://github.com/open-telemetry/semantic-conventions/pull/2504#issuecomment-3059181296 it sounds like this will need some time to discuss

schema/instrumentation.json
```diff
@@ -88,6 +88,12 @@
             "type": "object",
             "additionalProperties": false,
             "properties": {
+                "known_methods": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
```
trask, line 96: maybe this should be split into client and server options

I could almost see this option not being needed on client spans since it is primarily needed for avoiding metric cardinality attacks on servers (but I see we've defined it on both client and server spans in semconv, I don't remember if we discussed this previously, @lmolkova?)

Issue #255, opened 2025-07-31
Markdown based specification


thompson-tomo:
I would like to have the specification described in markdown files which are generated automatically based upon the schema definition files.

After some research & testing it appears that https://github.com/adobe/jsonschema2md/ can be used and it won't need an additional runtime ie go/python.

The Hope is once the docs are of quality ie all titles are set, they can be published to openTelemetry.io to maximise visibility/understanding.

An example of using the tool can be seen at: https://github.com/thompson-tomo/opentelemetry-configuration/tree/feature/%23255_MarkdownDocs/docs/spec

jack-berg:
I've had the same idea for some time now! Recently merged a variety of PRs related to what I'm calling the [meta schema](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#meta-schema). Amongst the changes, we now generate [schema-docs.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md) from information sourced from both the meta schema and JSON schema.

PR #266, opened 2025-08-11
First attempt at generating markdown specification

thompson-tomo:
Progresses #255 

This introduces a markdown based version of the specification which is generated via tooling which consumes the schema definition.

### DIFF
--- a/Markdown
+++ b/Markdown
@@ -1,0 +1,1 @@
+jsonschema2md -d schema -o docs -e .json -x -
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -1,0 +1,155 @@
+# README
+
+## Top-level Schemas
+
+* [Declaritive Config - Resource](./resource.md) – `https://opentelemetry.io/otelconfig/resource.json`
+
+* [OpenTelemetryConfiguration](./opentelemetry_configuration.md) – `https://opentelemetry.io/otelconfig/opentelemetry_configuration.json`
+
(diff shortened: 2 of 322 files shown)


jack-berg:
> data-model.md - Shows the file definition

You mean the JSON schema definition itself?

> Profiles - folder to store all profile definitions 

Not familiar with what you mean by a profile. JSON schema uses the vocabulary of types and properties. E.g. [OpenTelemetryConfiguration](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#opentelemetryconfiguration) is type with properties `file_format`, `disabled`, etc.

I suspect by profile you mean type. Assuming this, you're advocating for an organization of one file per type, as opposed to a single large file with all types as currently modeled in `schema-docs.md`? Personally, I lean towards the higher information density of keeping everything in one file. Of course, there are usability issues when the file gets too large, but typically markdown renderers have tooling to assist with navigation, like github's outline view:

Screenshot 2025-11-18 at 2 59 32 PM

If you feel strongly about this, open an issue and I'm happy to change course if that's the consensus.

> Those profiles have an explicit stability and defined seperately meaning the core data model is reduced hence easier to Stabilize.

Yup we differentiate between stable / experimental types and properties using the convention documented [here](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#experimental-features):

- Experimental types start with `Experimental*`, e.g. `ExperimentalLoggerConfig`
- Experimental properties end with the suffix `*/development`, e.g. `.instrumentation/development`

We're allowed to make breaking changes to experimental types and properties after stabilization, but not to stable types and properties.

PR #276, opened 2025-08-21
fix link checker

zeitlinger:
lychee update broke the checker

it seems the PR was merged despite the failing test

See https://github.com/lycheeverse/lychee/issues/1729


codeboten:
Thanks for the fix!

PR #295, opened 2025-09-15
chore: pin lychee-action

codeboten:
This will prevent renovate from trying to update the lychee-action which is pinned to v2.4.1 as newer versions cause build failures with false positives.


codeboten:
> If there are security issues won't this mean we will miss upgrades that resolve them?

Dependabot will still warn us of security issues

MrAlias (approved):

PR #312, opened 2025-10-02
Introduce proper meta schema concept

jack-berg:
Meta schema is the evolution of `type_descriptions.yaml`, and holds all types of details that don't fit neatly into the JSON schema. 

Initial use cases include:

* Property descriptions. Previously done via `type_descriptions.yaml`, we should evolve this to more precisely model the semantics. I.e. break out high level description, SDK interpretation, default behavior.
* SDK extension plugin interfaces. Tracking which types can reference SDK extension plugin interfaces via the ComponentProvider mechanism.
* SDK implementation status. Track which types and properties are supported by which SDK implementations. **Update** I've sketched this out on a separate branch [here](https://github.com/jack-berg/opentelemetry-configuration/tree/language-implementation-status) (see generated markdown [here](https://github.com/jack-berg/opentelemetry-configuration/blob/language-implementation-status/schema-docs.md)). Will open a followup PR to keep things slightly more reviewable. 

I wrote a useful summary of the meta schema and associated tasks here: https://github.com/jack-berg/opentelemetry-configuration/blob/meta-schema/CONTRIBUTING.md#meta-schema

Please read while reviewing!

### DIFF
--- a/.github/config/lychee.toml
+++ b/.github/config/lychee.toml
@@ -17,6 +17,6 @@ exclude = [
     # excluding links to pull requests and issues is done for performance
     '^https://github.com/open-telemetry/opentelemetry-configuration/(issues|pull)/\d+$',
     # exclude localhost, file references from schema-docs.md
-    "^http://localhost.*",
-    "^file://.*",
+    "^http://localhost",
+    "file:///path/to/file.jsonl",
 ]


jack-berg:
@open-telemetry/configuration-approvers anybody interested in discussed / reviewing this? Its a decent amount of code, but the stakes are low because its "build" code, as opposed to part of the schema, examples, stability guarantees. My view is we should land this and iterate.

marcalff (commented):
LGTM.

marcalff (approved):

PR #339, opened 2025-10-24
Add schema for ProbabilitySampler

### LINKED ISSUE #315: ProbabilitySampler Configuration
With https://github.com/open-telemetry/opentelemetry-specification/pull/4627 merged, what should the declarative configuration look like for this new setting? 

The previous `trace_id_ratio_based` configuration uses `ratio` as the modifier. Is this still the best sub-config for the Probability Sampler?

https://github.com/open-telemetry/opentelemetry-configuration/blob/865123526289121e233b1a72b4fb36da5621b64e/examples/kitchen-sink.yaml#L741-L745

[The configuration section in the specification hasn't changed](https://github.com/open-telemetry/opentelemetry-specification/blob/fbcd7a3126a545debd9e6e5c69b7b67d4ef1c156/specification/trace/sdk.md?plain=1#L522-L528), with the exception of the `Trace ID Ratio Based` name becoming `Probability`.

anuraaga:
Fixes #315

The user interface for `ProbabilitySampler` is identical to the legacy `TraceIdRatioBasedSampler` so the schema is the same.

I didn't update any examples because the sampler is still Development level and isn't provided in most SDKs yet.


jack-berg (commented):
Its good to have an example in `kitchen-sink.yaml` even for experimental types. Samplers are a bit of an edge case right now because you can only assign one to `.tracer_provider`.

Also, we're trying to keep [type_descriptions.yaml](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/type_descriptions.yaml) up to date with descriptions describing the semantics / defaults of all properties. Currently, those descriptions manifest as comments automatically added to the example files, but with no surface in the examples, it seems kind of silly.

I'm proposing a better way to maintaining extra schema meta data in #312, which amongst other things adds build tooling guaranteeing that every property for every type has a description.

schema/tracer_provider.json
```diff
@@ -123,6 +126,15 @@
                 }
             }
         },
+        "ExperimentalProbabilitySampler": {
+            "type": ["object", "null"],
+            "additionalProperties": false,
+            "properties": {
+                "ratio": {
+                    "type": ["number", "null"]
```
jack-berg: Could add `"minimum": 0`, `"maximum": 1`: https://json-schema.org/understanding-json-schema/reference/numeric#range

Or maybe `"exclusiveMinimum": 0` since the spec says the minimum valid ratio is `2^-56`: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#probabilitysampler-sampler-configuration

jack-berg (approved):

PR #350, opened 2025-10-29
Track language implementation status in meta schema

jack-berg:
Followup to #312.

This expands the meta schema to include tracking for language implementation status. Notes:

We want it to be easy for language implementation maintainers to add / update implementation status. Therefore, each language has their own `meta_schema_language_{language}.yaml` file, with all types listed and kept in sync automatically through build tooling. I could have added a language implementation status to each of the meta schema type entries, but this would be very cumbersome for language maintainers as they would need to scroll through a huge file with lots of irrelevant information just to update their relevant bits.

We want it to be easy for end users to understand implementation status as a glance. End users are typically interested in a specific language. Therefore, the markdown generation has section for the status of each language implementation [like this](https://github.com/jack-berg/opentelemetry-configuration/blob/meta-schema-language-status/schema-docs.md#language-support-status).

We also want config project maintainers to be able to understand the implementation status of particular types / properties across all languages at a glance, for things like deciding whether to promote experimental types to stable. Therefore, the markdown generation has a cross-language support status collapsible section for each type. For example, see "Language support status" section in [SpanExporter](https://github.com/jack-berg/opentelemetry-configuration/blob/meta-schema-language-status/schema-docs.md#spanexporter-) type definition.

I think we'll need to consider / iterate what implementation status information we want to capture and how best to organize it maintainability. Here's what I sketched out for this PR:

```
latestSupportedFileFormat: 1.0.0-rc.1
typeSupportStatuses:
  - type: Aggregation
    status: supported
    unsupportedProperties: []
    notes: ""
```

Notes:

- `.latestSupportFileFormat` captures the top level file format the implementation supports
- `.typeSupportStatuses` has entries for each JSON schema type, automatically synced with build tooling
- `.typeSupportStatuses[].status` is an enum with values: `supported`, `unknown` (default until a maintainer updates it). What other options are there?
- `.typeSupportStatuses[].unsupportedProperties` captures any properties that are unsupported. This is only applicable when `status: supported`.
- `.typeSupportStatuses[].notes` open ended string field capturing any specific callouts about the implementation's support of the type.

WDYT? What information is missing? Is there a representation that would be easier for maintainers?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -294,7 +294,7 @@ Notes:
 
 ### `meta_schema_language_{language}.yaml`
 
-These files track language implementation status for a particular language. See [fix-meta-schema](#fix-meta-schema) for details on adding a new language.
+These files track language implementation status for a particular language. See [fix-meta-schema](#make-fix-meta-schema) for details on adding a new language.
 
 Content looks like:
 


marcalff:
RE:

> .typeSupportStatuses[].status is an enum with values: supported, unknown (default until a maintainer updates it). What other options are there?

I would like to propose the following enum values as well:

* not implemented: The property is not even parsed. For example, some experimental nodes in opentelemetry-cpp.
* ignored: The property is parsed but is not used in the underlying SDK. For example, opentelemetry-cpp will ignore limits, because this is not implemented: our spec matrix still has holes in it. Ignored imply this is a known limitation, which may change later to supported.
* not applicable: The property is not parsed or is not used in the underlying SDK, because it does not make sense for this SDK. For example, the java instrumentation is not applicable for opentelemetry-cpp. Not applicable is a final state, to differentiate from ignored. Likewise, I would expect the cpp instrumentation node to be not applicable for java, claiming supported is very misleading.

In short, when we know exactly the status for an entry which is not supported, it should not stay blank or unknown but be given a more descriptive status.

jack-berg:
Suggestion makes sense @marcalff.

With multiple statuses emerging, we'll want to expand the schema for `unsupportedProperties` to be more specific since we'll want to be able to express things like:

```
- type: Foo
  status: supported
  propertyOverrides:
    - property: bar
      status: not_implemented
    - property: baz
      status: ignored
```

The interpretation of this would be: The type `Foo` and all its properties are supported except `bar` which is not implemented yet, and `baz` which is ignored.

In this proposal, `propertyOverrides` is only meaningful if the type's top level `status: supported`. I.e. you can't have a type which is `not_implemented` which has properties which are `supported`.

CONTRIBUTING.md
```diff
@@ -249,13 +249,82 @@ make all
...
+    * `.typeSupportStatuses[].propertyOverrides[].property` the name of the property whose support status is overridden.
+    * `.typeSupportStatuses[].propertyOverrides[].status` the overridden support status. See enum options below.
+* Status enum options, applicable to `.typeSupportStatuses[].status`, `.typeSupportStatuses[].propertyOverrides[].status`:
+  * `unknown`: Language maintainer has not yet recorded a status.
+  * `suppported`: The type / property is supported by the language implementation.
+  * `not_implemented`: The type / property is not parsed / recognized by the language implementation because the concept is not yet implemented but should be eventually.
+  * `not_applicable`: The type / property is not parsed / recognized by the language implementation because the concept is not applicable. E.g. C++ specific instrumentation for Java. 
+  * `ignored`: The type / property is not parsed / recognized by the language implementation despite the concept being available in the language's programmatic configuration API.
```
jack-berg: Adding lightweight docs for the schema of the `meta_schema*.yaml` documents.

PR #357, opened 2025-10-31
Split out defaultBehavior and nullBehavior from description in the meta schema

jack-berg:
Currently, all the semantics of properties are bundled up in the `[].properties[].description` field of `meta_schema_types.yaml`. This includes a terse description, various details about allowable values and semantics, and default behavior. 

We [schema modeling guidance](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#required-and-null-properties) which states that the description MUST include details about the semantics when the property is omitted / null.

These details are an important part of the stability contract and for ensuring consistency across language implementations. 

However, enforcement of this guidance is very difficult / error prone, and as we'll see in this PR, there are many instances where we have failed to provide those details.

This PR adds two new fields to the meta schema:

- `defaultBehavior` describes the behavior when a property is omitted (and in some cases when it is null). Build tooling enforces that this field is present if a property is not required, and that the field is not present if a property is required.
- `nullBehavior` describes the behavior when a property is null (if `defaultBehavior` is not defined). This allows us to define separate semantics for when a field is omitted entirely vs. present but with the value set to null. Build tooling enforces that this field is present if a property is required and is nullable. It can optionally be added to any non-required field.

If approved, I will go back in a followup PR and replace all the `TODO` values with proper descriptions.

Ironing out these semantics now will help ensure we have a more robust first stable release.

### DIFF
--- a/.github/config/lychee.toml
+++ b/.github/config/lychee.toml
@@ -19,4 +19,7 @@ exclude = [
     # exclude localhost, file references from schema-docs.md
     "^http://localhost",
     "file:///path/to/file.jsonl",
+    "file:///var/log/logs.jsonl",
+    "file:///var/log/metrics.jsonl",
+    "file:///var/log/traces.jsonl",
 ]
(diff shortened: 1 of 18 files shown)


jack-berg:
Superseded by #426.

PR #361, opened 2025-10-31
Add schema/meta_schema_language_cpp.yaml

marcalff:
Add coverage for the C++ language.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -141,7 +141,7 @@ logger_provider:
         # Configure exporter.
         exporter:
           # Configure exporter to be OTLP with file transport.
-          # This type is in development and subject to breaking changes in minor versions.
+          # This property is experimental and subject to breaking changes in minor versions.
           otlp_file/development:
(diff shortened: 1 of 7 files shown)


schema/meta_schema_language_cpp.yaml
```diff
@@ -56,6 +56,10 @@ typeSupportStatuses:
     status: supported
     notes: ""
     propertyOverrides: []
+  - type: ConsoleMetricExporter
```
jack-berg: @marcalff I merged some other PRs that created conflicts with this. I went ahead and resolved, but have made an assumption here c++ supports (or will imminently) support the new console metric exporter type.

Let me know and I'll get this merged.

marcalff: Yes @jack-berg , C++ will do that very soon.

https://github.com/open-telemetry/opentelemetry-cpp/blob/930b694af4f4c96f98d85514313d1dbaa8ed4b06/sdk/src/configuration/configuration_parser.cc#L608

Ok to merge.

PR #362, opened 2025-10-31
Add ConsoleMetricExporter

marcalff:
Fixes #242

When `PushMetricExporter` is a `console`,
add the following attributes:

* `temporality_preference`
* `default_histogram_aggregation`

PR #364, opened 2025-11-01
Update java implementation status

jack-berg:
Did a thorough pass on the java implementation and updated current implementation status.

### DIFF
--- a/schema-docs.md
+++ b/schema-docs.md
@@ -1330,9 +1330,9 @@ Usages:
 
 | Property | [cpp](#cpp) | [go](#go) | [java](#java) | [js](#js) |
 |---|---|---|---|---|
-| `endpoint` | not_implemented | unknown | unknown | unknown |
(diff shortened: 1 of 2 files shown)


MrAlias (approved):
As a configuration SIG member I can say the syntax looks good. I'd like to have another Java SIG member give this a review for content before merging.

jack-berg:
Will reach out to the java SIG. TBH, its a tough ask because a thorough review requires the reviewer to go through the implementation with a fine tooth comb like I needed to do to open this PR. I favor rubber stamps for this type of thing, and followup fixes if / when discovered.

jaydeluca (approved):

jack-berg:
Ah good catch @zeitlinger! I think I made an error merging conflicts from #362 since I called out support for specific properties while indicating that ConsoleMetricExporter itself was unsupported. Fixed in #380.

PR #365, opened 2025-11-01
breaking: View selector and stream should be required

jack-berg:
A view without a selector or stream is nonsensical.


marcalff:
Makes sense.

C++ already treat these as required, sorry I missed this in the schema (the yaml parser is hand written, not generated)

codeboten (commented):
This should be marked as breaking change correct?

codeboten:
We mark things as breaking in the changelog, but we should probably be better about updating the changelog in the PR itself rather than putting this responsibility on the release process

codeboten (approved):

PR #366, opened 2025-11-01
Make implementation status notes optional

jack-berg:
Implementation status notes are not used right now and add a lot of clutter to the `meta_schema_language_{language}.yaml` documents. 

This PR makes them optional.

### DIFF
--- a/schema/meta_schema_language_java.yaml
+++ b/schema/meta_schema_language_java.yaml
@@ -120,30 +120,27 @@ typeSupportStatuses:
     status: ignored
     propertyOverrides: []
   - type: ExperimentalLanguageSpecificInstrumentation
-    status: unknown
+    status: supported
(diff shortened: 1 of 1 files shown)


marcalff:
> We've had two implementation statuses reported and no notes recorded. Let's delete this and de-clutter the implementation status YAML files. Can restore later if a use case emerges.

In case of `translation_strategy` for prometheus, I have a case where not all enum values are supported.

I did not have time to provide details yet, but notes can be helpful for some edge cases.

jack-berg:
Good feedback. Will adjust to make it an optional addition, instead of automatically included on every type.

jack-berg:
Updated to retain implementation status notes but make them optional.

marcalff (approved):
LGTM

PR #376, opened 2025-11-05
Update prometheus exporter properties to match the spec

zeitlinger:
Prom spec update: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md#configuration

- without_target_info has added
- without_scope_info has a new meaning - it affects scope labels now (before: scope metric)


schema/meta_schema_language_go.yaml
```diff
@@ -159,7 +159,11 @@ typeSupportStatuses:
   - type: ExperimentalPrometheusMetricExporter
     status: unknown
     notes: ""
-    propertyOverrides: []
+    propertyOverrides:
+      - property: without_scope_info
+        status: unknown
+      - property: without_target_info
+        status: unknown
```
jack-berg: #nit: Technically unnecessary since the entire ExperimentalPrometheusMetricExporter has status unknown

schema/meter_provider.json
```diff
@@ -189,6 +189,9 @@
                 "without_scope_info": {
                     "type": ["boolean", "null"]
                 },
+                "without_target_info": {
+                    "type": ["boolean", "null"]
+                },
                 "with_resource_constant_labels": {
```
jack-berg: This appears to have been removed from the spec. Should remove, but can open a new issue to track rather than blocking this PR.

schema/meter_provider.json
jack-berg: Unrelated: enum type should be broken out as a dedicated type. 

https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#schemas-and-subschemas

schema/meter_provider.json
jack-berg: #404

PR #390, opened 2025-11-06
Add schema for composite sampler

### LINKED ISSUE #340: CompositeSampler configuration
There are currently Development level specs for a CompositeSampler which composes ComposableSamplers while handling trace state. 

https://opentelemetry.io/docs/specs/otel/trace/sdk/#compositesampler

4 of them are the composable version of existing samplers, AlwaysOn, AlwaysOff, Probability, ParentBased. The impression I get, though I could be wrong, is the long term goal is to transparently swap the existing samplers with the composable ones - they behave similarly while additionally sending trace state so are an additive change.

Until then, I propose a global flag on `sampler` itself, `composite: true`. If set, the sampler is `CompositeSampler` and any configuration below uses the composable version rather than the old version.

For composite-only samplers like `ComposableRuleBased`, probably we could make it an error to define that without defining `composite: true`, alternatively we could implicitly enable composite if it's present but an error seems clearer.

anuraaga:
Fixes #340

Essentially proposal from here but left out `delegate:`
https://github.com/open-telemetry/opentelemetry-configuration/issues/340#issuecomment-3461020987

Didn't actually agree on concensus in the issue so happy to make any changes, just wanted to keep the ball rolling.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,26 @@ The schema is modeled using JSON schema [draft 2020-12](https://json-schema.org/
 
 This is reflected in top level schema documents by setting `"$schema": "https://json-schema.org/draft/2020-12/schema"`.
 
+## What you see is what you get
+
+The schema semantics should follow a "what you see is what you get" (or WYSIWYG) philosophy. Another way to frame this is that implementations should minimize the amount of magic which occurs as a result of the absence of an optional property.
+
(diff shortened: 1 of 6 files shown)


schema/tracer_provider.json
```diff
@@ -89,6 +92,28 @@
...
+        "ExperimentalComposableSampler": {
+            "type": "object",
+            "additionalProperties": {
+                "type": ["object", "null"]
+            },
+            "minProperties": 1,
+            "maxProperties": 1,
+            "properties": {
+                "always_off": {
+                    "$ref": "#/$defs/AlwaysOffSampler"
```
jack-berg: Consistency with ExperimentalComposableParentBasedSampler would suggest separate ExperimentalComposableAlwaysOn/ExperimentalComposableAlwaysOff sampler as defined in the spec, but it seems silly at the same time.

jack-berg (approved):

Issue #393, opened 2025-11-07
Better project tooling for snippets


jack-berg:
Up till now, we use `kitchen-sink.yaml` as a blunt instrument to:

- Demonstrate the full configuration surface area of the schema
- Validate that the schema works as we expect (i.e. `make validate-examples`)

This is results in a variety of problems:

- `kitchen-sink.yaml` is huge!
- Despite our best effort, `kitchen-sink.yaml` can't express all the configurable surface area. For example, see #390 using bits of the parent based sampler configuration, which has 5 configurable sampler properties, to express a new sampler. 
- `kitchen-sink.yaml` configuration is often nonsensical, as a result of the effort to document all the surface area.
- `kitchen-sink.yaml` is annotated with comments generated from the JSON schema and meta schema, which is good for consistency, but limits our ability to convey bits of information about a specific example to users.

I think what we really need is:

- A directory of snippets, e.g. `/snippets`
- Where each file in `/snippets` contains a narrow demonstration of a particular piece of the larger schema
- With build tooling which validates that each file adheres to the particular type it is demonstrating, leveraging a file naming convention of the form `{type}_{snippet_description}.yaml` with build tooling validating according to the `{type}` extracted from the file name.
- Where snippets do not contain generated comments from the JSON schema and meta schema, but instead only include ad-hoc comments where needed
- The generated [schema-docs.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md) could leverage these snippets to detect usages of each type, and either link over to those snippets or directly embed their contents in the markdown.
- And where this tooling ultimately allows `kitchen-sink.yaml` to be deleted, as it is essentially decomposed into a bunch of smaller focussed pieces.

jack-berg:
With #397 merged, I'll be chip away at converting `kitchen-sink.yaml` to targeted snippets and ultimately delete it.

PR #397, opened 2025-11-10
Snippets

jack-berg:
First take on #393.

Details:

- Introduces a new `/snippets` directly with files named `<JsonSchemaType>_<snake_case_snippet_description>.yaml`
- Each file is a fully valid `OpenTelemetryConfiguration` type, but is meant to emphasize a particular snippet of a specific type, indicated by the `<JsonSchemaType>` in the file naming convention
- A new task `make validate-snippets` validates each snippet file against `OpenTelemetryConfiguration`, and the snippet being emphasized (as indicated by `# SNIPPET_START`) against its corresponding JSON schema type
- Snippets may include comments which describe a very narrow use case being demonstrated.
- The `make generate-markdown` task outputs snippets in `schema-docs.md`. For example, I've added a snippet demonstrating how to use a view to change the default histogram bucket boundaries for a specific instrument via a view. See screenshot below, or [generated markdown here](https://github.com/jack-berg/opentelemetry-configuration/blob/snippets/schema-docs.md#view-):

Screenshot 2025-11-10 at 11 46 04 AM

If we like this direction, we can open followup PRs which chip away at `kitchen-sink.yaml`, and delete it.

### DIFF
--- a/.github/workflows/build-check.yaml
+++ b/.github/workflows/build-check.yaml
@@ -25,8 +25,8 @@ jobs:
     - name: validate example
       run: make validate-examples
 
-    - name: meta schema
-      run: make all-meta-schema
+    - name: make all
+      run: make all
 
     - name: check for diff
       run: |
@@ -38,7 +38,7 @@ jobs:
         then 
           echo "No diff detected."
         else 
-          echo "Diff detected - did you run 'make all-meta-schema'?"
+          echo "Diff detected - did you run 'make all'?"
           echo $(git diff --cached --name-only)
           echo $(git diff --cached)
           exit 1


jack-berg:
@marcalff I've come around to the approach you've suggested, pushed a commit reflecting it, and updated the PR description.

A typical snippet now looks like:

File contents at `./snippets/Sampler_parent_based.yaml`:

```yaml
file_format: "1.0-rc.2"

tracer_provider:
  processors:
    - simple:
        exporter:
          console:
  # SNIPPET_START
  sampler:
    parent_based:
      root:
        trace_id_ratio_based:
          ratio: 0.0001
      remote_parent_sampled:
        always_on:
      remote_parent_not_sampled:
        probability/development:
          ratio: 0.01
      local_parent_sampled:
        always_on:
      local_parent_not_sampled:
        always_off:
```

- The build tooling will validate the entire file contents against the root `OpenTelemetryConfiguration` type. This means we often have to include some irrelevant extra bits to pass validation (e.g. `.tracer_provider.processors` here). But in exchange, implementations can now easily use the directly of snippets as tests.
- The file is named <JsonSchemaType>_<snake_case_description>`, so for this example:
  - JSON schema type: `Sampler`
  - Description (converted to title case): Parent Based
- The `# SNIPPET_START` indicates the start of the point of interest. The tooling will:
  - Identify this and for all the lines after it, strip whitespace equivalent to what proceeds `# SNIPPET_START` (in this case, two spaces will be stripped from following lines)
  - Take the resulting blob of content, and validate it against `<JsonSchemaType>` from the file name
- Markdown generation includes the portion after `# SNIPPET_START`, and provides a link back to the snippet source file so users can see the snippet in the context of the larger schema.

Pretty cool, I think. Should be very useful for users and maintainers alike.

PR #400, opened 2025-11-10
WIP: experiment with json5

codeboten:
This tries using json5 as the source of truth for the schema files. This allows us to use multi lines for descriptions, and a slightly less verbose json syntax. locally i'm using a json5 utility to generate the .json schema.

### DIFF
--- a/examples/kitchen-sink.yaml
+++ b/examples/kitchen-sink.yaml
@@ -459,34 +459,50 @@ meter_provider:
         exporter:
           # Configure exporter to be console.
           console:
-            # TODO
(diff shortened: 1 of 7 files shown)


codeboten:
> maybe use yaml-> json instead?

i was trying to stay as close to the resulting schema as possible with https://github.com/open-telemetry/opentelemetry-configuration/pull/396 and @jack-berg suggested using json5 for the support of multi line description and comments.

codeboten:
Closing this PR, i don't think adding json5 makes things easier for maintainers, contributors, or end users here

PR #407, opened 2025-11-14
Demonstrate using tooling to produce JSON schema files enriched with descriptions

jack-berg:
Alternative to #400 which uses tooling to produce JSON schema files enriched with descriptions from the meta schema. 

Main advantage here is that we can include descriptions which are a combination of bits of information from the meta schema:

- `description` from `meta_schema_types.yaml`
- `isSdkExtensionPlugin` from `metah_schema_types.yaml`
- Default and null behavior (see #357)
- (Possibly) snippets (see #397)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "scripts": {
     "generate-descriptions": "node scripts/generate-descriptions.js",
     "generate-markdown": "node scripts/generate-markdown.js",
-    "fix-meta-schema": "node scripts/fix-meta-schema.js"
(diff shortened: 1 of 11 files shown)


codeboten (commented):
I would suggest that if this is the desired direction, we should remove the json schema and represent the data in yaml as the source of truth.

One problem with moving away from the source schema altogether is that the yaml and tooling then needs to be able to represent any relationship that json schema does, which may cost more down the road if we find new ways to use the schema that we have yet to implement.

Having both yaml and json schema be combined to produce another json makes it harder to make understand how this all works together.

PR #410, opened 2025-11-14
Add schema for rule based sampler

### LINKED ISSUE #341: ComposableRuleBasedSampler configuration
There is a new rule-based sampler in the spec

https://opentelemetry.io/docs/specs/otel/trace/sdk/#composablerulebased

Unlike every other sampler, this one needs special care in configuration since we can't "just model args as JSON" - one of the args is an arbitrary predicate function. So we need to define what types of predicates should be supported specifically by declarative configuration. For starters, I suggest the following

- `attribute_value`: takes an attribute key and value and performs an exact match
- `attribute_pattern`: takes an attribute key and regex pattern and performs a regex match
- `span_kind`: takes a span kind enum name and performs an exact match
- `is_root`: no parameters, matches if the span is a root span.

We can allow all of these to be defined at the same time to perform an `AND` composition (it should probably be an error to have both attribute specs though). If no predicate is specified, it is catch-all - in practice, most rule sets will almost always have this in the end because the default for a non-match is to drop.

The rule would contain the predicates, and `sampler`, with the latter being identical to the current `Sampler` schema.

Example
```
sampler:
  rule_based:
    rules:
      - attribute_value:
          key: http.route
          value: /health
        sampler:
          always_off: 
      - span_kind: CLIENT
        is_root: true
        sampler:
          always_off:
      - sampler: # The default sampler, parent-based always-on
```

Let me know how that sounds. /cc @jack-berg

anuraaga:
Fixes #341 

Given the complexity of the schema I had to take many liberties with the details - let me know any issues or possible improvements. /cc @jack-berg

### DIFF
--- a/opentelemetry_configuration.json
+++ b/opentelemetry_configuration.json
@@ -7,7 +7,7 @@
   "properties": {
     "file_format": {
       "type": "string",
-      "description": "The file format version.\nThe yaml format is documented at\nhttps://github.com/open-telemetry/opentelemetry-configuration/tree/main/schema\nProperty is required and must be non-null.\n"
+      "description": "The file format version.\nRepresented as a string including the semver major, minor version numbers (and optionally the meta tag). For example: \"0.4\", \"1.0-rc.2\", \"1.0\" (after stable release).\nSee https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md for more details.\nThe yaml format is documented at https://github.com/open-telemetry/opentelemetry-configuration/tree/main/schema\nProperty is required and must be non-null.\n"
     },
     "disabled": {
       "type": [
@@ -2148,10 +2148,6 @@
         "console": {
           "$ref": "#/$defs/ConsoleExporter",
           "description": "Configure exporter to be console.\nIf omitted, ignore.\n"
-        },
-        "zipkin": {
-          "$ref": "#/$defs/ZipkinSpanExporter",
-          "description": "Configure exporter to be zipkin.\nIf omitted, ignore.\n"
         }
       }
     },
@@ -2437,30 +2433,6 @@
         "attribute_keys": {
           "$ref": "#/$defs/IncludeExclude",
           "description": "Configure attribute keys retained in the resulting stream(s).\nIf omitted, all attribute keys are retained.\n"
-        }
-      }
-    },
-    "ZipkinSpanExporter": {
-      "type": [
-        "object",
-        "null"
-      ],
-      "additionalProperties": false,
-      "properties": {
-        "endpoint": {
-          "type": [
-            "string",
-            "null"
-          ],
-          "description": "Configure endpoint.\nIf omitted or null, http://localhost:9411/api/v2/spans is used.\n"
-        },
-        "timeout": {
-          "type": [
-            "integer",
-            "null"
-          ],
-          "minimum": 0,
-          "description": "Configure max time (in milliseconds) to wait for each export. \nValue must be non-negative. A value of 0 indicates indefinite.\nIf omitted or null, 10000 is used.\n"
         }
       }
     }
--- a/schema/tracer_provider.yaml
+++ b/schema/tracer_provider.yaml
@@ -444,10 +444,6 @@ $defs:
         $ref: common.yaml#/$defs/ConsoleExporter
         description: Configure exporter to be console.
         defaultBehavior: ignore
-      zipkin:
-        $ref: "#/$defs/ZipkinSpanExporter"
-        description: Configure exporter to be zipkin.
-        defaultBehavior: ignore
     isSdkExtensionPlugin: true
   SpanLimits:
     type: object
@@ -525,28 +521,6 @@ $defs:
         description: Configure a simple span processor.
         defaultBehavior: ignore
     isSdkExtensionPlugin: true
-  ZipkinSpanExporter:
-    type:
-      - object
-      - "null"
-    additionalProperties: false
-    properties:
-      endpoint:
-        type:
-          - string
-          - "null"
-        description: |
-          Configure endpoint.
-        defaultBehavior: http://localhost:9411/api/v2/spans is used
-      timeout:
-        type:
-          - integer
-          - "null"
-        minimum: 0
-        description: |
-          Configure max time (in milliseconds) to wait for each export. 
-          Value must be non-negative. A value of 0 indicates indefinite.
-        defaultBehavior: 10000 is used
   ExperimentalTracerConfigurator:
     type:
       - object
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ for general information about the project.
 
 The following rules are enforced when modeling the configuration schema.
 
-These rules must be enforced when making changes to the schema; however, changes to the rules are permitted. In other words, the [stability definition](./README.md#stability-definition) does not apply to the rules: so long as the stability definition is not violated for existing properties, the modeling rules guiding new properties may change.
+These rules must be enforced when making changes to the schema; however, changes to the rules are permitted. In other words, the [versioning policy guarantees](VERSIONING.md#guarantees-and-allowed-changes) do not apply to the rules: so long as the versioning policy is not violated for existing properties, the modeling rules guiding new properties may change.
 
 ### Which JSON schema version?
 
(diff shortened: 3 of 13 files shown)


schema/meta_schema_types.yaml
```diff
@@ -1264,3 +1266,48 @@
...
+      description: INTERNAL, an internal span.
+    - enumValue: PRODUCER
+      description: PRODUCER, a producer span.
+    - enumValue: SERVER
+      description: SERVER, a server span.
+  isSdkExtensionPlugin: false
+- type: ExperimentalComposableRuleBasedSampler
+  properties:
+    - property: rules
+      description: The rules for the sampler, matched in order. Each rule can have multiple match conditions - the sampler will be applied if all match. If no conditions are specified, the rule matches all spans that reach it.
```
jack-berg: Probably worth describing that if no rules are matched, the sampler uses a non-sampling intent: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#composablerulebased

schema/meta_schema_types.yaml
```diff
@@ -1264,3 +1266,48 @@
...
+  isSdkExtensionPlugin: false
+- type: ExperimentalComposableRuleBasedSamplerRule
+  properties:
+    - property: attribute_patterns
+      description: Patterns to match against a single attribute.
+    - property: attribute_values
+      description: Values to match against a single attribute.
+    - property: is_root
```
jack-berg: Would be cool to be able to differentiate between the trace root and the root span within a service's boundaries (sometimes called the local root).

schema/tracer_provider.json
```diff
@@ -210,9 +248,42 @@
...
+        },
+        "ExperimentalComposableRuleBasedSamplerRuleAttributePatterns": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "key": {
+                    "type": "string"
+                },
+                "patterns": {
+                    "$ref": "common.json#/$defs/IncludeExclude"
```
jack-berg: 👍

schema/tracer_provider.json
```diff
@@ -210,9 +248,42 @@
...
+            "additionalProperties": false,
+            "properties": {
+                "key": {
+                    "type": "string"
+                },
+                "values": {
+                    "type": "array",
+                    "minItems": 1,
+                    "items": {
+                        "type": "string"
```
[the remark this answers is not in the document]
jack-berg: > I think most array attributes are semantically an AND, i.e. the entire gen ai prompt, the entire cmd args.

Not sure I'm following, but I'll demonstrate my expectation by example. Given an attribute: `foo: ["bar", "baz", "qux"]`, and a matcher like:
```
attribute_patterns:
  key: foo
  patterns:
    included:
      - ba*
```

I would expect a match even tho `qux` doesn't match the pattern. I.e. any match semantics. 

> Should I start by narrowing the comment to say this currently only supports string keys to push back both of the above?

You can do that, but I'm just as happy saying that all attributes are converted to strings before comparison, and waiting for feedback indicating problems with that simplifying assumption. I imagine `http.response.status_code` being a popular problem case for a "we only accept string attributes" interpretation.

I think your experience with OTTL is probably relevant here so will defer to you 🙂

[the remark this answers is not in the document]
anuraaga: > I would expect a match even tho qux doesn't match the pattern. I.e. any match semantics.

Yeah I was just thinking of what array attributes I've seen before, probably the genai wasn't a good example since it's a complex attribute. Yeah actually ANY sounds better if matching process.args for whatever reason.

> I think your experience with OTTL is probably relevant here so will defer to you

I think yaml is particularly troubling here due to the optional quotes. But yeah, string keys was overly strict since it's easy to convert numbers / bool to string. Clarified that any attribute will be compared as a string

Makefile
```diff
@@ -26,7 +26,7 @@ update-file-format:
 	done
 
 .PHONY: fix-meta-schema
-fix-meta-schema:
+fix-meta-schema: compile-schema
```
anuraaga: For whatever reason I had the habit of `make all-meta-schema`, which would then skip the yaml -> json step causing all sorts of mayhem.

Probably the generation script can use a safety guard to fail when the JSON is empty

examples/kitchen-sink.yaml
```diff
@@ -289,8 +289,45 @@ tracer_provider:
...
+                  key: http.path
+                  patterns:
+                    included:
+                      - /internal/*
+                    excluded:
+                      - /internal/special/*
+                # Configure sampler when matched.
+                sampler:
+                  # Configure sampler to be always_on if matched.
+                  always_on: # Configure a sampling rule matching root spans with CLIENT span kind.
```
PeterF778: Split this line to start the comment in the next line.

examples/kitchen-sink.yaml
```diff
@@ -289,8 +289,45 @@ tracer_provider:
...
+            rules:
+              - attribute_values:
+                  key: http.route
+                  values:
+                    - /healthz
+                    - /livez
+                # Configure sampler when matched.
+                sampler:
+                  # Configure sampler to be always_off if matched.
+                  always_off: # Configure a sampling rule matching http.path attribute patterns.
```
PeterF778: Splitting this line will improve readability.

examples/kitchen-sink.yaml
```diff
@@ -289,8 +289,45 @@ tracer_provider:
...
+              - attribute_patterns:
+                  key: http.path
+                  patterns:
```
PeterF778: I wonder if "patterns:" can be eliminated from the schema. The context is already about patterns.

PeterF778:
I think this PR strikes a good balance between ergonomics and flexibility. My guess is that in practice it will cover more than 90% of cases. And it can be easily extended, if needed, by adding more _ExperimentalComposableRuleBasedSamplerRules_ to match span names, for example, or to provide building blocks for complex predicates (`or`, `not`, etc.).

schema/tracer_provider.yaml
```diff
@@ -22,6 +22,38 @@ properties:
...
+      - CLIENT
+      - PRODUCER
+      - CONSUMER
+    enumDescriptions:
+      INTERNAL: INTERNAL, an internal span.
+      SERVER: SERVER, a server span.
+      CLIENT: CLIENT, a client span.
+      PRODUCER: PRODUCER, a producer span.
+      CONSUMER: CONSUMER, a consumer span.
+  SpanParent:
```
anuraaga: #427 made me realize perhaps the rules need to capture the existing knobs of the `ParentBasedSampler` and went with this approach. @PeterF778 let me know if this makes sense

opentelemetry_configuration.json
```diff
@@ -520,6 +520,113 @@
...
+        },
+        "span_kinds": {
+          "type": "array",
+          "minItems": 1,
+          "items": {
+            "$ref": "#/$defs/SpanKind"
+          },
+          "description": "The span kinds to match. If the span's kind matches any of these, it matches."
+        },
+        "parent": {
```
PeterF778: I found that a useful predicate for rule based sampling is `no_local_parent`, which nicely identifies service head spans. This, of course, in addition to `is_trace_root`.  I realize that the proposed `parent` rule is expressive enough, but it is more verbose.

schema/tracer_provider.yaml
anuraaga: Thanks, definitely should have been `ExperimentalSpanParent`

opentelemetry_configuration.json
```diff
@@ -2085,6 +2209,19 @@
...
+      "type": [
+        "string",
+        "null"
+      ],
+      "enum": [
+        "NONE",
+        "REMOTE_NOT_SAMPLED",
+        "REMOTE_SAMPLED",
+        "LOCAL_NOT_SAMPLED",
+        "LOCAL_SAMPLED"
```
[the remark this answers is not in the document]
anuraaga: Thanks - I removed the sampled distinction

opentelemetry_configuration.json
anuraaga: I went with keeping the parent enum since then we don't have to worry about "what if both are set" type of situations. I guess in terms of verbosity, `parent: LOCAL` is similar to `no_local_parent: true`? Possibly helped by removing the SAMPLED state from the enum, let me know if it looks fine now

schema/tracer_provider.yaml
```diff
@@ -22,6 +22,34 @@ properties:
 required:
   - processors
 $defs:
+  SpanKind:
+    type:
+      - string
+      - "null"
+    enum:
```
jack-berg: FYI, I'm advocating we standardize on lower_snake_case for enum values in #445

jack-berg: Applies to `ExperimentalSpanParent` as well.

Issue #422, opened 2025-11-19
Release multiple complete schema files


thompson-tomo:
Currently we release 1 single collection of json schema files (8 files) however to improve shareability and distribution it would be useful if we opted to release a complete file from from local references.

I am mindful that this would lead to a growth of the file size and to accommodate for this I would propose to release 3 complete files:
- Core: only defines schema upto the extension points. This is useful when vendors/languages want to define their own collection of extensions based on what they support.
- OTEL: core + extensions defined in otel spec.
- Complete: OTEL + any other extensions which has a configuration definition.

thompson-tomo:
@jack-berg I see #425 as an example of the complete file which is 1 of the 3 files mentioned.

PR #425, opened 2025-11-20
Merge output schema into single file and check into git

jack-berg:
Possibly solves #422. I alluded to this in a comment [here](https://github.com/open-telemetry/opentelemetry-configuration/pull/411#discussion_r2536061198).

There are a variety of advantages to publishing a single single JSON schema file and checking it into git:

- Maintaining source schema in YAML has its advantages, but one big disadvantage is that it obfuscates how changes actually manifest in the resolved JSON schema. Like `schema-docs.md`, this helps make the impact of changes obvious.
- It can be annoying to configure JSON schema tooling to be able to properly resolve `$ref`s across files. With everything in one file, configuring tooling couldn't be easier.
- Without the compiled JSON schema committed to git, its more difficult to test out changes in an implementation. I need to update my implementation to point to a ref, download `opentelemetry-configuration` at the ref, run `make compile-schema`, and have my implementation point to `./schema_out`. Much easier if every ref has the fully compiled schema checked into source code in a single file.
- Makes publishing easier because we don't need to compile / zip / attach the schema to the release. Its always available at the ref.

It makes schema changes a bit noisier in terms of diff, but If we can tolerate keeping `schema-md.doc` checked into git, we can tolerate this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ for general information about the project.
 
 ## Schema modeling rules
 
-The following rules are enforced when modeling the configuration schema. 
+The following rules are enforced when modeling the configuration schema.
 
 These rules must be enforced when making changes to the schema; however, changes to the rules are permitted. In other words, the [stability definition](./README.md#stability-definition) does not apply to the rules: so long as the stability definition is not violated for existing properties, the modeling rules guiding new properties may change.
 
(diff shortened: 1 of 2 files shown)


Makefile.common
jack-berg: Reverted these changes: https://github.com/open-telemetry/opentelemetry-configuration/pull/411#discussion_r2539107071

With the compiled schema checked in, `validator` is no longer dependent on the root `compile-schema` target and all the changes to accommodate that are no longer required.

schema-docs.md
jack-berg: Only changes to this file are the removal of redundant `$defs` in the rendering of each type's concrete schema.

marcalff:
> It makes schema changes a bit noisier in terms of diff, but If we can tolerate keeping `schema-md.doc` checked into git, we can tolerate this.

I agree.

Even when the full schema in json and the schema-docs.md are generated, it makes sense to have them checked in in git.

People consuming this needs to see the full schema.

marcalff (approved):

PR #426, opened 2025-11-20
Add defaultBehavior and nullBehavior to source schema

jack-berg:
#357, but adjusted for the current organization where `meta_schema_types.yaml` has been merged.

We have [schema modeling guidance](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#required-and-null-properties) which states that the description MUST include details about the semantics when the property is omitted / null.

These details are an important part of the stability contract and for ensuring consistency across language implementations. 

However, enforcement of this guidance is very difficult / error prone, and as we'll see in this PR, there are many instances where we have failed to provide those details.

This PR adds two new fields to the source schema YAML which are not part of standard JSON schema:

- `defaultBehavior` describes the behavior when a property is omitted (and in some cases when it is null). `make compile-schema` enforces that this field is present if a property is not required, and that the field is not present if a property is required.
- `nullBehavior` describes the behavior when a property is null (if `defaultBehavior` is not defined). This allows us to define separate semantics for when a field is omitted entirely vs. present but with the value set to null. `make compile-schema` enforces that this field is present if a property is required and is nullable. It can optionally be added to any non-required field.

The compiled schema `opentelemetry_configuration.json` enriches property `description`s with the information from these fields, so that the information can be propagated to comments in generated code.

`schema_docs.md` displays this information in the property table for each type. For example:

Screenshot 2025-11-20 at 4 46 34 PM

Where we already have default semantics defined in descriptions, this PR extracts those into these new fields. There are many many places where we did not follow our schema modeling guidance and `defaultBehavior` is left as a TODO. Defining this behavior is left for future PR(s).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,12 +30,12 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@fdbfb4d2750291e159f0156def62b853c2798ca2 # v4.31.5
+        uses: github/codeql-action/init@fe4161a26a8629af62121b670040955b330f9af2 # v4.31.6
         with:
           languages: go
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@fdbfb4d2750291e159f0156def62b853c2798ca2 # v4.31.5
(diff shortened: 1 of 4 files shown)


opentelemetry_configuration.json
jack-berg: Note how the descriptions include default / null semantics as a suffix.

PR #435, opened 2025-11-26
Default behavior todos

jack-berg:
Followup to #426.

Common themes:

- For all properties in a SDK extension plugin, set `defaultBehavior: ignore`. 
- For properties with complex types, where those types have defaults of their own for each property, set `defaultBeahvior: default values as described in <type> are used`

There weren't too many default behaviors that were completely missing. Mostly just dots that needed to be connected.

There was only one case where the defaults were unspecified and not obvious - the jaeger remote sampler. I made `endpoint` and `initial_sampler` required, and set the `defaultBehavior: 60000 is used` for `internal` (the default in java but not defined in the spec). This case isn't too important because we marked jaeger remote sampler as experimental in #363.

### DIFF
--- a/opentelemetry_configuration.json
+++ b/opentelemetry_configuration.json
@@ -25,7 +25,7 @@
     },
     "attribute_limits": {
       "$ref": "#/$defs/AttributeLimits",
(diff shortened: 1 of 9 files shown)


jack-berg:
Closing in favor of #439.

PR #445, opened 2025-12-02
Add schema modeling guidance for enum values as lower_snake_case

jack-berg:
Realized we don't consistently model enum value types when reviewing this [spec PR](https://github.com/open-telemetry/opentelemetry-specification/pull/4723#discussion_r2577482207). 

Ideally, we would have case insensitive enums, but JSON schema doesn't support those without resorting to using regex patterns.

Choosing snake case because we already have modeling guidance saying that property names are snake case.


MrAlias (commented):
Should the values be case-insensitive? Do we want to distinguish between `info` and `INFO`?

jack-berg:
See #449.

Notably:

> Advantage of upper snake case is that enums stick out very clearly.
> Disadvantage is that more types are impacted:
> - lower_snake_case impacts 2 types, both of which are experimental
> - UPPER_SNAKE_CASE impacts 7 types, one of which is experimental, 6 of which are on stack to stabilize

codeboten (commented):
If you can address the conflicts and update the description for the enum, i think we can move this forward

PR #446, opened 2025-12-02
Change instrument type refs from observable to asynchronous

jack-berg:
The spec uses both "observable" and "asynchronous" to describe metric instruments which are invoked by callback.

* [observable](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-specification%20observable&type=code) is used 6 times
* [asynchronous](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-specification+asynchronous&type=code) is used 36 times

I think I carried "observable" over from [the java implementation](https://github.com/open-telemetry/opentelemetry-java/blob/76c622cb0b7334d621a2f635daaf4d4ee2fb8d74/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/InstrumentType.java#L18-L20). We can keep it, but if we prefer better alignment with the spec now is the time to change it.

### DIFF
--- a/opentelemetry_configuration.json
+++ b/opentelemetry_configuration.json
@@ -231,7 +231,7 @@
           ],
           "minimum": -10,
           "maximum": 20,
-          "description": "TODO\nIf omitted or null, TODO.\n"
+          "description": "Configure the max scale factor.\nIf omitted or null, TODO.\n"
         },
         "max_size": {
           "type": [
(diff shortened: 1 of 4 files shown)


jack-berg:
I am fine with keeping it as observable. It would be a breaking change which I'd probably characterize as [not low impact](https://github.com/open-telemetry/opentelemetry-configuration/pull/456/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR5-R7), so maybe the bar needs to be higher to justify.

codeboten:
Maybe this is a consistency we address once implementations have renamed their apis from observable to asynchronous 🤷🏻

PR #447, opened 2025-12-02
Update .log_level to be enum

### LINKED ISSUE #387: Consider making log_level experimental
The [`.log_level`](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/opentelemetry_configuration.json#L14-L16) property semantics are pretty loose: 

> Configure the log level of the internal logger used by the SDK. If omitted, info is used.

I believe we talked in the past about this being language specific. If so, we should call this out. If there is a known enum of values (I doubt all language maintainers will agree), we should list the values.

jack-berg:
Resolves #387.

The spec [OTEL_LOG_LEVEL](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration) env var describes itself as an enum, but never states the set of enum values.

This is a mistake we shouldn't repeat in declarative config. The lack of stricter semantics for the env var was probably from the desire to avoid bikeshedding about what those enum values should actually be. But we've come a long way from those early days, and have a stable log data model and proto which include severity number definition:

- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-severitynumber
- https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L87-L113

This is the obvious choice to use for `.log_level`. 

Failing to do this would perpetuate language inconsistency around `OTEL_LOG_LEVEL` to the next generation of configuration, which would be a shame.

### DIFF
--- a/.github/workflows/build-check.yaml
+++ b/.github/workflows/build-check.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: check out code
-      uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
     - name: install dependencies
       run: npm install
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout repository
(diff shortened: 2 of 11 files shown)


MrAlias (approved):

jack-berg:
> Is the choice to use this over https://github.com/open-telemetry/opentelemetry-configuration/pull/445?

This is independent of #445. This is about changing `.log_level` to be an enum. #445 is about consistency about enum casing. If we merge #445, there will be merge conflicts with this.

PR #456, opened 2025-12-04
Prepare for 1.0-rc.3 release

jack-berg:
~~Couple of straggler PRs we should get resolved:~~

~~- #433 not important for stability, but seems reasonable and there is strong interest~~
~~- #445 or #449 to decide on standard case for enum values~~
~~- #446 not strictly necessary but won't be able to do it later~~
~~- #450 easy and uncontroversial~~
~~- #451 nice to have for better compiled descriptions~~


jack-berg:
Should merge #460

MrAlias (approved):

PR #461, opened 2025-12-11
Tweak rule based sampler

jack-berg:
Nothing important for rc3.


opentelemetry_configuration.json
anuraaga: Thanks for the help @jack-berg! I agree with your https://github.com/open-telemetry/opentelemetry-java/pull/7861#discussion_r2611833407 missed that case when reasoning out the use cases. Should we have a `predicates` array of these that are ANDed?

PR #472, opened 2025-12-18
Extract otlp_http, otlp_grpc exporter examples from kitchen sink to snippets

jack-berg:
Related to #393

### DIFF
--- a/snippets/OtlpGrpcExporter_logs_kitchen_sink.yaml
+++ b/snippets/OtlpGrpcExporter_logs_kitchen_sink.yaml
@@ -1,4 +1,4 @@
-file_format: "1.0-rc.2"
+file_format: "1.0-rc.3"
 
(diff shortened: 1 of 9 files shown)


snippets/OtlpGrpcExporter_logs_kitchen_sink.yaml
```diff
@@ -0,0 +1,20 @@
+file_format: "1.0-rc.2"
+
+tracer_provider:
```
codeboten: the name of the file and the contents dont align here and in traces_kitchen_sink

jack-berg: bah

jack-berg: fixed

PR #474, opened 2025-12-20
Fix prom exporter without scope info

zeitlinger:
As specified in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md#configuration

### DIFF
--- a/schema/meter_provider.yaml
+++ b/schema/meter_provider.yaml
@@ -19,7 +19,7 @@ properties:
   exemplar_filter:
     $ref: "#/$defs/ExemplarFilter"
     description: |
-      Configure the exemplar filter. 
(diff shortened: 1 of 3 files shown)


schema/meter_provider.yaml
```diff
@@ -257,7 +257,7 @@ $defs:
           - boolean
           - "null"
         description: |
-          Configure Prometheus Exporter to produce metrics without a scope info metric.
+          Configure Prometheus Exporter to produce metrics without scope info labels.
```
jack-berg:
```suggestion
          Configure Prometheus Exporter to produce metrics without scope labels.
```

PR #481, opened 2025-12-29
Extract limits examples from kitchen sink to snippets

jack-berg:
Related to #393

### DIFF
--- a/schema-docs.md
+++ b/schema-docs.md
@@ -682,8 +682,8 @@ Snippets:
 
 [Snippet Source File](./snippets/CardinalityLimits_kitchen_sink.yaml)
 ```yaml
-# specify .default to set cardinality limits for all instruments, or specify instrument-specific limits. Instrument-specific limits take priority over .default.
-default: 2000
+# specify .default to set cardinality limits for all instruments, or specify instrument-specific limits. Instrument-specific limits take priority over .default. In this contrived example, the cardinality limits of each instrument type are set to 2000, overriding the value from .default.
+default: 100
 counter: 2000
 gauge: 2000
 histogram: 2000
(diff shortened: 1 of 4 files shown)


examples/kitchen-sink.yaml
```diff
@@ -10,17 +10,11 @@
 file_format: "1.0-rc.3"
 disabled: false
 log_level: info
-attribute_limits:
-  attribute_value_length_limit: 4096
-  attribute_count_limit: 128
```
marcalff: To consider, also create a snippet for the top level `attributes_limits` node, so that ultimately every node has an example in the doc.

jack-berg: Yeah I've been going back and forth about this.. Now that we have a tool for snippets, what should our goal be? 

One direction:
- At least one snippet for every type, illustrating the "kitchen sink" of available properties for that type
- Other snippets for popular use cases of the type as needed

But this has some downsides:
- There's a lot of types, so there will be a lot of snippets! Some of them may start to feel repetitive. 
- Snippets for different extension points end up scatter around the docs. For example, right now we have snippets for [OtlpHttpExporter](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#otlphttpexporter-) and [ExperimentalOtlpFileExporter](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#experimentalotlpfileexporter), but none for [SpanExporter](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#spanexporter-). It may be more useful to aggregate the snippets for the built-in extension plugins in the SDK extension plugin type. I.e. keep all the built-in span exporter snippets in the in SpanExporter type, keep all the built-in sampler snippets in the Sampler type. This would allow readers to see the different type of exporter configuration at a glance, without having to jump around a bunch of places. But it would mean a more nuanced snippet policy than "one snippet for every type".

PR #489, opened 2026-01-05
Extract console exporter examples from kitchen sink to snippets, delete kitchen-sink.yaml

jack-berg:
Related to #393


marcalff (approved):
LGTM

PR #556, opened 2026-03-04
Clarify pattern matching is case-sensitive

trask:
Related to https://github.com/open-telemetry/opentelemetry-java/issues/8137

cc @SylvainJuge


jack-berg (approved):

SylvainJuge (approved):

zeitlinger (approved):
Looks good — making case-sensitivity explicit is the right call. Consistent with how wildcards already behave, and SylvainJuge's concern is resolved in the comments.

PR #600, opened 2026-04-22
ci: update to rust version of flint

zeitlinger:
This updates the existing link linting to the current Rust-based `flint` release.

What changes in this PR:
- install `flint` from `github:grafana/flint = "0.20.3"`
- replace the old contrib-hosted `lint:links` shell tasks with `flint run`
- keep the existing Lychee configuration under `.github/config`
- enable `check_all_local = true` so local links are checked across the repo
- preserve full-repo link checking on pushes by keeping `mise run lint --full` on the non-PR workflow path

Why this stays small:
- this repo does not currently carry Markdown lint tooling to migrate here
- the goal is only to replace the old link-check task path with the released Rust `flint` path

Verification:
- `mise run lint`
- `mise run lint --full`


codeboten (commented):
the changes look ok. i'm not super familiar w/ flint... was there a discussion about switching to it for link checking somewhere?

zeitlinger:
> the changes look ok. i'm not super familiar w/ flint... was there a discussion about switching to it for link checking somewhere?

It's been mostly within the Java ecosystem so far - tldr is that I created flint because  the link checking kept getting more complicated.

Here you can see that all other repos that have previously used those shell scripts for link checking have now switched to flint: https://github.com/grafana/flint/issues/140#issuecomment-4329330325

codeboten (approved):

PR #617, opened 2026-05-12
Clarify AttributeNameValue description, fix related tooling

### LINKED ISSUE #613: what is the intent for *empty* resource attribute values?
**Update**: See my comment just below. I'm not clear what the intent is for *empty* resource attribute values.

---

`value` property of `AttributeNameValue` allows null, but description includes "must be non-null"

https://github.com/open-telemetry/opentelemetry-configuration/blob/2a4cebcb0cb5c7c1545d7bf6733131b73d1cd705/opentelemetry_configuration.json#L146-L171

At a guess, is this because `isNullable = types.includes('null')` in scripts/source-schema.js doesn't handle the `oneOf` usage?
https://github.com/open-telemetry/opentelemetry-configuration/blob/2a4cebcb0cb5c7c1545d7bf6733131b73d1cd705/scripts/source-schema.js#L153-L164

If it is confirmed that the "must be non-null" here is incorrect, then I can take a stab at a PR.

jack-berg:
Resolves #613.

The generated description of AttributeNameValue indicates that value is "Property is required and must be non-null."

But this isn't true. It is required, but it can be present and nullable.

The docs generation tooling has bug where it incorrectly generates the description when `"null"` type is in a `oneOf`.

This PR fixes the tooling, and adds a `nullBehavior`, resulting in a new value description of "Property must be present, but if null the entry is ignored."


trentm (approved):
Thanks!

codeboten (approved):
👍🏻

PR #624, opened 2026-05-19
add IdGenerator definition to TracerProvider

### LINKED ISSUE #70: Add IdGenerator in the trace_provider
The specification defines IdGenerator as an extension point:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#id-generators

The yaml for `tracer_provider:` should probably allow to define which id generator to use, something like:

```
tracer_provider:
  id_generator:
    # SDK extension point
    my_custom_id_generator:
```

Found while prototyping for opentelemetry-cpp

MikeGoldsmith:
## Summary

Adds `IdGenerator` to the schema as an SDK extension plugin under `TracerProvider`, with a single built-in `random` for the spec-default random TraceId/SpanId generation.

Modeled consistently with existing SDK extension plugins (`Sampler`, `SpanExporter`, `SpanProcessor`). Treated as stable (no `/development` suffix) — IdGenerator is a stable spec extension point in [trace/sdk.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#id-generators), and the plugin shape (name + opaque properties bag) has nothing in flux. The property is optional and falls back to the SDK's random default when omitted.

Closes #70. Unblocks opentelemetry-js#6616 and similar work in other languages (originally surfaced from opentelemetry-cpp prototyping).

## Notes

- Built-in name chosen: `random` (matches Java autoconfigure's `OTEL_TRACES_ID_GENERATOR=random` convention). The spec doesn't standardize a name.
- `RandomIdGenerator` has no configuration properties — the spec doesn't define any.
- Language status files have new types stubbed as `unknown`; the inline property-level status for `id_generator` inherits `TracerProvider`'s overall status per the existing tooling behavior. Maintainers can add `propertyOverrides` in follow-up PRs.

## Test plan

- [x] `make` passes (compile-schema, validate-examples, validate-snippets, fix-language-implementations, generate-markdown)
- [x] `opentelemetry_configuration.json` contains `IdGenerator` and `RandomIdGenerator` `$defs`
- [x] `schema-docs.md` lists `IdGenerator` under SDK Extension Plugins
- [x] `IdGenerator_custom.yaml` snippet validates against the compiled schema
- [ ] Reviewer confirms `random` is acceptable vs. `default` / `default_random`

### DIFF
--- a/schema/tracer_provider.yaml
+++ b/schema/tracer_provider.yaml
@@ -16,6 +16,11 @@ properties:
     description: |
       Configure the sampler.
     defaultBehavior: parent based sampler with a root of always_on is used
+  id_generator:
+    $ref: "#/$defs/IdGenerator"
+    description: |
+      Configure the trace and span ID generator.
+    defaultBehavior: an ID generator which randomly generates TraceIds and SpanIds is used
   tracer_configurator/development:
     $ref: "#/$defs/ExperimentalTracerConfigurator"
     description: |
@@ -396,6 +401,25 @@ $defs:
         $ref: "#/$defs/ExperimentalComposableRuleBasedSampler"
         description: Configure sampler to be rule_based.
         defaultBehavior: ignore
+  IdGenerator:
+    type: object
+    additionalProperties:
+      type:
+        - object
+        - "null"
+    minProperties: 1
+    maxProperties: 1
+    properties:
+      random:
+        $ref: "#/$defs/RandomIdGenerator"
+        description: Configure the ID generator to randomly generate TraceIds and SpanIds (spec default).
+        defaultBehavior: ignore
+    isSdkExtensionPlugin: true
+  RandomIdGenerator:
+    type:
+      - object
+      - "null"
+    additionalProperties: false
   SimpleSpanProcessor:
     type: object
     additionalProperties: false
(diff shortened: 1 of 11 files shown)


MikeGoldsmith:
ping @open-telemetry/configuration-approvers

codeboten (approved):

schema/tracer_provider.yaml
```diff
@@ -16,6 +16,11 @@ properties:
     description: |
       Configure the sampler.
     defaultBehavior: parent based sampler with a root of always_on is used
+  id_generator:
+    $ref: "#/$defs/IdGenerator"
+    description: |
+      Configure the trace and span ID generator.
+    defaultBehavior: an ID generator which randomly generates TraceIds and SpanIds is used
```
jack-berg: Maybe be explicit about which id generator is used and rely on the user to go look up the semantics of that. Else, we end up describing random id generator in two places. 
```suggestion
    defaultBehavior: RandomIdGenerator is used
```

jack-berg (approved):
Just a nit on the description. Thanks!

MikeGoldsmith:
Thanks both 👍🏻

PR #648, opened 2026-06-05
chore: update github actions via zizmor

codeboten:
Ran zizmor --fix=all . on the repo to automatically fix some of the issues it found.

### DIFF
--- a/.github/workflows/build-check.yaml
+++ b/.github/workflows/build-check.yaml
@@ -15,6 +15,8 @@ jobs:
     steps:
     - name: check out code
       uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      with:
+        persist-credentials: false
 
     - name: install dependencies
       run: npm install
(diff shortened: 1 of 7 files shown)


.github/workflows/build-check.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
     steps:
     - name: check out code
       uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      with:
+        persist-credentials: false
```
jack-berg, line 19: Not sure what this does. We don't have any secrets in this whole repo.

Can't hurt I suppose

.github/workflows/validator-release.yml
```diff
@@ -192,7 +195,7 @@ jobs:
       - id: cargo-dist
         shell: bash
         run: |
-          dist build ${{ needs.plan.outputs.tag-flag }} --output-format=json "--artifacts=global" > dist-manifest.json
+          dist build ${NEEDS_PLAN_OUTPUTS_TAG_FLAG} --output-format=json "--artifacts=global" > dist-manifest.json
```
jack-berg, line 200: Is this equivalent?

**Edit** Oh I see NEEDS_PLAN_OUTPUTS_TAG_FLAG is set below. Let's bring that env block up above run to make it clear.

jack-berg (approved):

MrAlias (approved):

.github/workflows/validator-release.yml
codeboten: done

PR #655, opened 2026-06-12
Deprecate `MetricProducer.opencensus`

Krishnachaitanyakc:
Follow-up to [spec PR #5138](https://github.com/open-telemetry/opentelemetry-specification/pull/5138) (merged 2026-06-12), which deprecated OpenCensus compatibility requirements in the specification (removal no earlier than June 2027).

## What this PR does

1. Adds `deprecated: true` (JSON Schema 2020-12 [annotation](https://json-schema.org/draft/2020-12/json-schema-validation#name-deprecated)) to `$defs.MetricProducer.properties.opencensus` and `$defs.OpenCensusMetricProducer` in [`schema/meter_provider.yaml`](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/meter_provider.yaml).
2. Adds a `**Deprecated**` banner to the `description` text of each entry — this is what surfaces the deprecation in `schema-docs.md`'s `## MetricProducer` property table, since the markdown generator does not render `deprecated:` separately.
3. Regenerates `opentelemetry_configuration.json` and `schema-docs.md` via `make all`.
4. Adds a `## Unreleased` CHANGELOG entry.

Non-breaking per [VERSIONING.md L65](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md#L65): "Adding, removing, or modifying `deprecated` annotation" is explicitly allowed in minor versions.

## Out of scope

- **Not removing** any entry. SDKs may continue honoring existing user configs.
- **Not deprecating `MetricProducer`** (the type) — it continues hosting future producers (see #576 for `prometheus`).
- **Not changing per-language `status` values** — the status enum has no `deprecated` value.

## Downstream

[opentelemetry-js#6712](https://github.com/open-telemetry/opentelemetry-js/pull/6712) (merged 2026-06-04) just wired up `OpenCensusMetricProducer` in `@opentelemetry/sdk-node`; it already emits `diag.warn` for unknown producer keys, so this is runtime-safe. @trentm's [comment](https://github.com/open-telemetry/opentelemetry-js/pull/6712#issuecomment-4614872906) anticipates this follow-up.

cc @jack-berg, @codeboten, @marcalff, @MikeGoldsmith, @maryliag, @dashpole.


maryliag (approved):

marcalff (commented):
Thanks for the pr.

Looks ok to deprecate, the main concern here is the wording:

> This entry will be removed no earlier than June 2027

comes from the spec, and refers to the specification itself: the OpenCensus specification will be removed after June 2027, but this does not mean every SIG SDK will drop OpenCensus by then.

The yaml node is likely to stay for SDK that still wants to support OpenCensus, even after June 2027.

Beside, removing a yaml node is a breaking change for the file format, so it is likely to stay for a really long time, with SDK accepting the syntax but raising an error at runtime.

When I read "This entry will be removed no earlier than June 2027", it gives me the impression the yaml node will be removed after June 2027, which is probably not what we want.

@open-telemetry/configuration-maintainers WDYT ?

Krishnachaitanyakc:
Thanks @marcalff, fair point. Dropped the 2027 removal part and kept "deprecated, SDKs MAY continue, new configs SHOULD NOT use".

marcalff (approved):
LGTM, thanks for the fix.

MrAlias (approved):

trentm (approved):

codeboten (approved):
👍🏻

jack-berg:
FYI, this is the first time we've deprecated something. The way it has been done establishes a precedent. Is everyone ok with this being the way we deprecate things going forward?

codeboten:
@jack-berg i think this approach makes sense, with the intent on keeping any deprecated fields in the schema until a major revision

jack-berg:
See followup in https://github.com/open-telemetry/opentelemetry-configuration/pull/659

PR #666, opened 2026-06-23
feat: add experimental view_matching_mode to meter provider

dashpole:
Implements composable View matching configuration in support of open-telemetry/opentelemetry-specification#5173.

Adds experimental meter_provider.view_matching_mode/development enum property, with ['independent' default, 'composable'] as values


examples/otel-sdk-config.yaml
```diff
@@ -72,6 +72,21 @@ meter_provider:
             timeout: 10000
             temporality_preference: cumulative
             default_histogram_aggregation: explicit_bucket_histogram
+  view_matching_mode/development: composable
+  views:
```
jack-berg: `/examples` is fully curated examples we intend for people to use as a starting point and without modification

This belongs as a [snippet](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#snippets)

schema/meter_provider.yaml
```diff
@@ -1,6 +1,15 @@
 type: object
 additionalProperties: false
 properties:
+  view_matching_mode/development:
```
jack-berg: Please model enums as top level types, and reference. Also, please add [enumDescriptions](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#json-schema-source-and-output) for description of each enum entry.

Example enum definition: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/common.yaml#L249-L302
And example usage: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/opentelemetry_configuration.yaml#L18-L22

examples/otel-sdk-config.yaml
dashpole: moved to a snippet

schema/meter_provider.yaml
dashpole: done, and added enum descriptions

schema-docs.md
jack-berg, line 1: Hey schema-docs.md is moved. Can you revert these changes, merge main and rerun the build?

jack-berg (commented):
Content looks good. Still need to fix `schema-docs.md`, mark "ready for review", and get more approvals on the spec PR. But setting aside that, the schema changes look good here.

schema-docs.md
dashpole: addressed in 0f7a679

snippets/MeterProvider_composable_views.yaml
```diff
@@ -0,0 +1,18 @@
+file_format: "1.1"
+
+meter_provider:
+  # SNIPPET_START
+  readers:
+    - periodic:
+        exporter:
+          console:
+  views:
+    - selector:
```
jack-berg, line 15: If you like you can leave comments describing what's interesting about this usage. Somethign like:

```
# By default, an instrument named "my_instrument" would match both of the registered views and output two metrics, one with description "first view description", the other with the original description and a cardinality limit of 1000. By setting the matching mode to composable, the matching views are merged together and one metric output with description "first view description" AND cardinality limit of 1000. 
```

jack-berg (approved):

snippets/MeterProvider_composable_views.yaml
dashpole: added in 32f537e

codeboten (approved):

PR #677, opened 2026-06-29
Group all patch and minor updates into weekly update

jack-berg:
Trying to reduce the clutter of renovate updates by grouping minor updates into a weekly update with the patches.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -7,12 +7,12 @@
   ],
   "packageRules": [
     {
-      "groupName": "all patch versions",
-      "matchUpdateTypes": ["patch"],
-      "schedule": ["before 8am every weekday"]
+      "groupName": "non-major dependencies",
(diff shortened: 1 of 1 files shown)


MrAlias:
From my experience in other SIGs that have tried this it resulted in large PRs that failed CI due to one or two dependencies and resulted in no updates being performed for months at a time. We have since moved away from this.

jack-berg:
We still do it in opentelemetry-java.

From my perspective, if I'm going to have to intervene in dependency updates, I'd rather it be once a week all at once.

Don't feel strongly about this though.

codeboten:
dependency updates have not caused a lot of issues with this repo, I would be ok with grouping them, but if folks feel more comfortable with more specific groups, we could start with something like this: https://github.com/open-telemetry/opentelemetry-configuration/pull/685

jack-berg:
Let's go with #685 for now. Can come back to it if needed

PR #685, opened 2026-07-06
chore: group github action updates

codeboten:
These are grouped in the collector repos and we don't often have issues with them, i recommend doing the same here


MrAlias (approved):

marcalff:
No strong opinion here one way or another.
Ok to merge for me.

Issue #689, opened 2026-07-12
Naming and stability are different concepts but they are represented like one single thing


ocelotl:
## Summary

Today, experimental parts of the schema are marked by appending a `/(development|alpha|beta)` suffix to the **name** of a property or enum value (e.g. `detection/development`, `otlp_file/development`, `no_translation/development`), per [VERSIONING.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md). This is a clear, explicit convention and it does its job of making instability visible.

I'd like to open a discussion about whether that visibility could be achieved a different way. My thinking is that the current approach combines two concerns that might be cleaner to keep separate: the _identity_ of a thing (its name) and the _stability_ of that thing (metadata about it, which changes over time). I'll lay out the reasoning and a couple of options below — very open to the possibility that there are constraints or history I'm missing.

## The observation: naming and stability are different concerns

A name identifies something; its stability is a fact _about_ it that changes over time. When the two are combined, a change in stability implies a change in identity.

A small analogy to make the intuition concrete. Imagine two buildings named like this:

- `EmpireState`
- `BurjKhalifa/being_renovated`

The second name feels a little off: "being renovated" is a temporary status, not part of what the building is. And when the renovation finishes, the building would need to be **renamed** to `BurjKhalifa`, so anything that referred to the older name would need updating.

The `/development` suffix has a similar shape. `detection/development` is "the `detection` property, currently unstable." When it stabilizes, it's renamed to `detection`, and downstream references to the older name need to change. That churn is a natural consequence of the stability signal living in the name.

## A practical wrinkle in code generation

Because the schema is used to generate code across many languages, property and enum names tend to become identifiers (fields, members, keys). The `/` character isn't a legal identifier character in most target languages, so each generator needs a step to translate a name like `detection/development` into something like `detection_development`. Each language handles this independently.

We recently ran into this in Python: [open-telemetry/opentelemetry-python#5395](https://github.com/open-telemetry/opentelemetry-python/issues/5395). A valid config like:

```yaml
resource:
  attributes:
    - name: service.name
      value: ${OTEL_SERVICE_NAME:-foo}
  detection/development:
    detectors:
      - service: {}
      - process: {}
      - host: {}
```

produced:

```
Resource.__init__() got an unexpected keyword argument 'detection/development'
```

The generated field was `detection_development`, but the YAML key is `detection/development`, so the mapping between them has to be handled explicitly at parse time.

[...]

PR #691, opened 2026-07-12
Add check for non-identifier characters in schema names

### LINKED ISSUE #690: There is no guard against special characters being used in the schema
## Summary

I'd like to propose a schema modeling rule: **every property name and enum value in the configuration schema should consist only of ASCII letters, digits, and underscores, and should not begin with a digit** — i.e. match `^[A-Za-z_][A-Za-z0-9_]*$`.

The motivation is that these names become **identifiers** in the many languages we generate code for, and this character set is a safe common denominator that is a valid identifier almost everywhere. Constraining names this way means no code generator has to sanitize or translate names, which removes a whole class of potential bugs — not just the one we have today, but ones we might otherwise introduce in the future.

## Background: the concrete case

The immediate motivation is the `/(development|alpha|beta)` suffix used to mark experimental properties and enum values (e.g. `detection/development`). The `/` isn't a legal identifier character in most languages, so each generator has to translate `detection/development` into something like `detection_development`, and the mapping has to be handled at parse time. We hit a bug from exactly this in Python — [open-telemetry/opentelemetry-python#5395](https://github.com/open-telemetry/opentelemetry-python/issues/5395):

```
Resource.__init__() got an unexpected keyword argument 'detection/development'
```

I've opened a separate discussion about that specific convention (whether stability should be expressed as metadata rather than in the name). **This issue is intentionally narrower and more general:** regardless of how the stability question is resolved, it seems worth having a rule about which characters are allowed in names at all, so we don't reintroduce the same shape of problem later with a different character.

## Why a general rule, not just a fix for `/`

Today `/` is the only non-identifier character in use. But nothing currently prevents a future property or enum value from using `.`, `-`, `:`, a space, or another character that is fine in JSON/YAML but problematic as an identifier. If that happened, we'd rediscover the same class of issue — a value that's valid in the schema and in a config file but that every language's code generator has to special-case, with each getting the chance to get it wrong independently.

A single up-front constraint avoids that entire category:

- **Portable by construction** — names that match `^[A-Za-z_][A-Za-z0-9_]*$` are valid identifiers in effectively every target language, so no per-language sanitization is needed.
- **No hidden mapping** — the key in a config file, the name in the schema, and the generated identifier can all be the same string. Less to get wrong, less to document.
- **Forward-looking** — it's a guardrail that prevents the next occurrence, not just a fix for the current one.

## Proposal

Add a rule to the [schema modeling rules](https://github.com/open-telemetry/opentelemetry-configuration?tab=readme-ov-file#schema-modeling-rules), roughly:

> Property names and enum values MUST match `^[A-Za-z_][A-Za-z0-9_]*$` (ASCII letters, digits, and underscores; not beginning with a digit). This keeps them usable as identifiers across the languages that generate code from the schema.

It could be enforced mechanically in the `compile-schema` / validation tooling by walking the compiled schema and checking every property key and enum value against the pattern, so a violating name fails the build rather than being discovered downstream.

A couple of details worth deciding as part of this:

- **Scope.** The rule is most important for property names and enum values, since those become identifiers. `$defs` type names are already effectively identifiers (PascalCase), so they'd comply as-is, but we could state the rule for them too for consistency.
- **Case.** Property names and enum values are currently snake_case, so a stricter `^[a-z_][a-z0-9_]*$` would also fit today's usage if we prefer to encode the casing convention at the same time. I'd lean toward the more permissive letters-and-digits form for the identifier rule and keeping casing as a separate style guideline, but either works.
- **Existing exceptions.** The only current violations are the `/(development|alpha|beta)` names. Adopting this rule would depend on resolving those (see the companion discussion), so the two are related — this rule is the general principle, and the stability discussion is the specific case that currently conflicts with it.

## Related

- **#5395** (opentelemetry-python) — the downstream bug from the `/` character.
- **#145** ("Should schema modeling rules be normative?") — relevant to whether a rule like this would be stated as MUST/SHOULD.
- Companion discussion on the `/development` stability convention (the specific case this general rule would cover).

## Closing thought

This is meant as a small, preventive guardrail rather than a criticism of any current naming. Even once the `/development` question is settled, a rule that keeps all names within a portable identifier character set seems like cheap insurance against re-encountering the same class of problem with some other character down the line. Interested in whether others see value in codifying it, and if so, at what strictness.

ocelotl:
Fixes #690.

## What

Adds a check that every **property name** and **enum/const value** in the compiled schema is a portable identifier, matching `^[A-Za-z_][A-Za-z0-9_]*$` (ASCII letters, digits, underscores; not starting with a digit).

- `scripts/check-schema-names.js` — walks the compiled `opentelemetry_configuration.json`, reporting any violating name with a JSON-path breadcrumb.
- `make check-schema-names` (npm script `check-schema-names`) — runs it. Included in `make all`, so it runs in the existing **Build Check** CI job with no workflow change.
- `CONTRIBUTING.md` — documents the rule under the schema modeling rules.

## Why

Property names and enum values become **identifiers** in the many languages that generate code from this schema. A character that isn't valid in an identifier forces every generator to sanitize/translate names independently, which is a recurring, per-language source of bugs — e.g. the `/` in `detection/development` caused open-telemetry/opentelemetry-python#5395.

Today `/` is the only offending character, but nothing guards against a future property using `.`, `-`, `:`, a space, etc. This is that guard: it prevents the whole class of problem up front, not just the current instance.

## Rollout (why it's a draft / warn-only)

The schema currently contains 18 names that violate the rule (the `/(development|alpha|beta)` stability suffix). So the check ships in **warn-only mode**: it prints violations but exits `0`, so it does not break the build. A single `ENFORCE` flag in the script flips it to a hard failure once those names are resolved (see the related discussion on expressing stability as metadata rather than in names).

Current output against `main`:

```
WARNING: found 18 name(s) that are not valid identifiers (must match /^[A-Za-z_][A-Za-z0-9_]*$/):

  - property "detection/development"  (at #.$defs.Resource.properties)
  - property "otlp_file/development"  (at #.$defs.SpanExporter.properties)
  - enum value "no_translation/development"  (at #.$defs.ExperimentalPrometheusTranslationStrategy.enum)
  ... (18 total)
```

## Open questions

- **Warn-only vs. enforce:** happy to keep warn-only, or gate enforcement on resolving the `/development` names.
- **Scope:** the check covers property names and enum/const values (the identifier-bearing names). `$defs` type names already comply (PascalCase); should the rule cover them explicitly too?
- **Strictness:** the rule allows uppercase for generality; property names/enum values are snake_case today, so a stricter `^[a-z_][a-z0-9_]*$` would also fit if we want to encode casing here rather than leaving it to the existing case rules.


marcalff:
> see the related discussion on expressing stability as metadata rather than in names

For reference, this is in #689

jack-berg (changes requested):
I want to consider this more closely before adding tooling that warns on the presence of suffixes which are the documented and recommended way to accomplish experimental vs. stable distinction.

jack-berg:
Upon further thought, I would be in favor moving forward with this and enforcing loudly with a build error if we could explicitly allow the currently intentional `/(development|alpha|beta)` suffixes.

If we change our stance on how we denote maturity, we can update the build tooling at the same time.

ocelotl:
> Upon further thought, I would be in favor moving forward with this and enforcing loudly with a build error if we could explicitly allow the currently intentional `/(development|alpha|beta)` suffixes.
> 
> If we change our stance on how we denote maturity, we can update the build tooling at the same time.

Added `/development` as an exception :v:

scripts/check-schema-names.js
jack-berg, line 1: Let's roll this into `compile-schema.js`. We already have various verification steps to ensure the types of schema cleanliness rules we enforce on top of JSON schema itself, and this is the same type of thing: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/scripts/compile-schema.js#L11-L15

"If you try to add a property with an invalid identifier, schema compilation fails"

scripts/check-schema-names.js
```diff
@@ -0,0 +1,83 @@
...
+
+import fs from "fs";
+import { schemaPath } from "./util.js";
+
+const IDENTIFIER = /^[A-Za-z_][A-Za-z0-9_]*$/;
+
+// The intentional `/development` maturity suffix that is allowed to trail an
+// identifier. If the stability-in-names convention changes, update this here
+// (see #689).
+const MATURITY_SUFFIX = /\/development$/;
```
jack-berg: Extend the comment to say that /alpha, /beta suffixes could also be included as allowed but are omitted because we don't have any instances of them at the moment.

CONTRIBUTING.md
```diff
@@ -56,6 +56,10 @@ Properties defined in the schema should be lower [snake case](https://en.wikiped
 
 [enum](https://json-schema.org/understanding-json-schema/reference/enum) values should be lower [snake case](https://en.wikipedia.org/wiki/Snake_case).
 
+### Property name and enum value character set
```
jack-berg, line 59: By convention, type names are CamelCase a-zA-Z0-9 today as well. Should probably formalize this and add verification for it as well. I'm happy to address separately, but you could also do it in this PR if you're interested. 

Oddly, we added a schema modeling rule indicating that [property names are lower_snake_case](https://github.com/open-telemetry/opentelemetry-configuration/blob/f1602adfdda3daa5ca978f131743691b73240133/CONTRIBUTING.md?plain=1#L51-L54), but never codified the convention for types.

scripts/check-schema-names.js
ocelotl: fixed

CONTRIBUTING.md
ocelotl: Good catch, added it.

scripts/check-schema-names.js
ocelotl: fixed

scripts/compile-schema.js
```diff
@@ -199,6 +221,26 @@ function noSubschemas(sourceSchemaType, messages) {
     });
 }
 
+function namesShouldBeValidIdentifiers(sourceSchemaType, messages) {
+    const isValidName = name => identifier.test(name.replace(maturitySuffix, ''));
+    sourceSchemaType.properties.forEach(property => {
+        if (!isValidName(property.property)) {
+            messages.push(`Property name '${property.property}' in ${sourceSchemaType.type} must match ${identifier} (optionally followed by a /development maturity suffix); it becomes an identifier in generated code. See #690.`);
```
jack-berg: #nit: I find the reference to the issue too verbose. 

```suggestion
            messages.push(`Property name '${property.property}' in ${sourceSchemaType.type} must match ${identifier} (optionally followed by a /development maturity suffix)`);
```

(Applies to the other error messages) as well.

CONTRIBUTING.md
```diff
@@ -56,6 +56,14 @@ Properties defined in the schema should be lower [snake case](https://en.wikiped
...
+
+### Type name case
+
+Type names (the top-level types defined under `$defs`, e.g. `BatchSpanProcessor`) should be [PascalCase](https://en.wikipedia.org/wiki/Camel_case): they must match `^[A-Z][A-Za-z0-9]*$`, a leading uppercase ASCII letter followed by ASCII letters and digits. Like property names and enum values, type names become identifiers in the languages that generate code from the schema. Experimental types are denoted with an `Experimental` prefix rather than a `/development` suffix. This rule is enforced by `make compile-schema`, which fails the build if you add a type with a name that is not PascalCase.
```
jack-berg: Keep it terse, link to VERSIONING.md as the source of truth.

Update the [relevant section in CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CONTRIBUTING.md#schema-validation) to indicate that this verification occurs in `make compile-schema`. 

```suggestion
Type names (the top-level types defined under `$defs`, e.g. `BatchSpanProcessor`) should be [PascalCase](https://en.wikipedia.org/wiki/Camel_case): they must match `^[A-Z][A-Za-z0-9]*$`. This facilitates usage as identifiers in generated code. Experimental types are [denoted with an `Experimental` prefix](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md#experimental-features).
```

CONTRIBUTING.md
```diff
@@ -56,6 +56,14 @@ Properties defined in the schema should be lower [snake case](https://en.wikiped
 
 [enum](https://json-schema.org/understanding-json-schema/reference/enum) values should be lower [snake case](https://en.wikipedia.org/wiki/Snake_case).
 
+### Property name and enum value character set
+
+Property names and enum values must match `^[A-Za-z_][A-Za-z0-9_]*$`: ASCII letters, digits, and underscores, not starting with a digit. These names become identifiers in the languages that generate code from the schema, so restricting them to this portable set means no code generator has to sanitize or translate a name (see [#690](https://github.com/open-telemetry/opentelemetry-configuration/issues/690)). The one exception is the intentional `/development` stability suffix (e.g. `detection/development`): a name is accepted when the part before that suffix is a valid identifier. This rule is enforced by `make compile-schema`, which fails the build if you add a property or enum value with an invalid identifier.
```
jack-berg:
```suggestion
Property names and enum values must match `^[A-Za-z_][A-Za-z0-9_]*$`. This facilitates usage as identifiers in generated code. Experimental properties are an exception, and are [denoted with a `/(development|alpha|beta)` suffix](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md#experimental-features) (e.g. `detection/development`).
```

scripts/compile-schema.js
```diff
@@ -2,6 +2,26 @@ import fs from 'fs';
 import {rootTypeName, schemaPath} from "./util.js";
 import {readSourceTypesByType} from "./source-schema.js";
 
+// Property names and enum values must be portable identifiers: ASCII letters,
+// digits, and underscores, not starting with a digit. These names become
+// identifiers in the languages that generate code from the schema, so
+// restricting them to this common set means no generator has to sanitize or
+// translate a name. See #690. The one exception is the intentional
+// `/development` maturity suffix (e.g. `detection/development`): a name is
+// accepted when the part before that suffix is a valid identifier.
```
jack-berg: I find the AIs are obnoxiously verbose in their comments. I instruct them "Omit comments unless the code is dense / complex enough to easily be misinterpretted without further explanation. Avoid having comment duplicating information. When comments are necessary, keep them terse."

```suggestion
```

If you agree, apply to the comments on the other regex as well.

jack-berg (approved):
Just nits.

PR #697, opened 2026-07-14
Add config for entity aware resource detectors

jack-berg:
Declarative config analog of `OTEL_EXPERIMENTAL_ENTITIES_ENALBED` as proposed in https://github.com/open-telemetry/opentelemetry-specification/pull/5210

Adds new `$.resource.detection/development.entities_enabled` property, defaulting to false. 

In the spec PR 5210, resource detectors are updated to always be emit entity aware resources. I.e. resources with entities attached. 

If `$.resource.detection/development.entities_enabled` is false, entity information is erased from resource detectors. If true, its retained.

### DIFF
--- a/schema/resource.yaml
+++ b/schema/resource.yaml
@@ -107,6 +107,17 @@ $defs:
           Configure resource detectors.
           Resource detector names are dependent on the SDK language ecosystem. Please consult documentation for each respective language. 
         defaultBehavior: no resource detectors are enabled
+      entities_enabled:
+        type:
+          - boolean
+          - "null"
+        description: |
+          Configure whether resource detectors emit entity-aware resources.
+
+          Resource detectors always produce entity-aware output internally. When this option is
+          not true, the SDK erases entity information from detector results before merging
+          them into the resource.
+        defaultBehavior: false is used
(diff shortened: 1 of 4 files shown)


codeboten (commented):
The change looks ok as proposed, one question comes to mind though. The default for the other `_enabled` configuration options (`scope_info_enabled` and `target_info_enabled/development`) is `true` if null or omitted, are we ok with the difference in behaviour here?

jsuereth (approved):

schema/resource.yaml
```diff
@@ -107,6 +107,17 @@ $defs:
...
+        type:
+          - boolean
+          - "null"
+        description: |
+          Configure whether resource detectors emit entity-aware resources.
+
+          Resource detectors always produce entity-aware output internally. When this option is
+          not true, the SDK erases entity information from detector results before merging
+          them into the resource.
+        defaultBehavior: false is used
```
jack-berg, line 120: Spec PR seems [poised](https://github.com/open-telemetry/opentelemetry-specification/pull/5147/changes#r3666938809) to change the default for entities to be enabled by default. 

Don't merge without updating accordingly.

PR #698, opened 2026-07-14
schema: add AlwaysRecord sampler

### LINKED ISSUE #83: Best way to support record only configuration
The OTel specification allows users to set a [recording only](https://github.com/open-telemetry/opentelemetry-specification/blob/700f5138f4b10c6e9d11b41153f6e164c61f581d/specification/trace/sdk.md?plain=1#L343) option for sampler configuration.  This option is not available via environment variables today: https://github.com/open-telemetry/opentelemetry-specification/blob/700f5138f4b10c6e9d11b41153f6e164c61f581d/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

Should this use-case be supported by the configuration file? If so would it be as simple as adding one more type of sampler here: https://github.com/open-telemetry/opentelemetry-configuration/blob/6215d43ee00c4ebe32ca8cbaf4c2807600b73945/schema/tracer_provider.json#L57

dlanov:
## Summary

Add declarative configuration schema support for the stable OpenTelemetry `AlwaysRecord` sampler.

This change:

- adds `sampler.always_record`
- adds the `AlwaysRecordSampler` schema type with a required wrapped `root` sampler
- adds a focused configuration snippet
- regenerates the compiled schema and language-support documentation
- marks language-specific support for the new sampler as unknown pending confirmation from language maintainers

`AlwaysRecord` converts `DROP` decisions from the wrapped sampler to `RECORD_ONLY`, allowing all spans to reach span processors while only sampled spans are sent to exporters.

Fixes #83

## Validation

- `make all`
- `git diff --check`
- all examples validate
- all snippets validate
- compiled JSON schema validates
- repeated generation produces no additional changes

### DIFF
--- a/schema/tracer_provider.yaml
+++ b/schema/tracer_provider.yaml
@@ -117,6 +117,10 @@ $defs:
         $ref: "#/$defs/AlwaysOnSampler"
         description: Configure sampler to be always_on.
         defaultBehavior: ignore
+      always_record:
+        $ref: "#/$defs/AlwaysRecordSampler"
+        description: Configure sampler to be always_record.
+        defaultBehavior: ignore
       composite/development:
         $ref: "#/$defs/ExperimentalComposableSampler"
         description: Configure sampler to be composite.
@@ -148,6 +152,20 @@ $defs:
       - object
       - "null"
     additionalProperties: false
+  AlwaysRecordSampler:
+    type:
+      - object
+    additionalProperties: false
+    properties:
+      root:
+        $ref: "#/$defs/Sampler"
+        description: |
+          Configure the wrapped sampler which provides the original sampling
+          decision that always_record modifies. DROP decisions are converted
+          to RECORD_ONLY, so spans are passed to processors without being
+          exported.
+    required:
+      - root
   ExperimentalJaegerRemoteSampler:
     type:
       - object
--- a/schema/tracer_provider.yaml
+++ b/schema/tracer_provider.yaml
@@ -161,9 +161,8 @@ $defs:
         $ref: "#/$defs/Sampler"
         description: |
           Configure the wrapped sampler which provides the original sampling
-          decision that always_record modifies. DROP decisions are converted
-          to RECORD_ONLY, so spans are passed to processors without being
-          exported.
+          decision that AlwaysRecord modifies. DROP decisions are converted
+          to RECORD_ONLY, allowing processors to see all spans without sending them to exporters.
     required:
       - root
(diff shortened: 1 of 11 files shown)


jack-berg:
Please mark "Ready for review" when you're ready to receive feedback and make progress towards merging 🙂

schema/tracer_provider.yaml
```diff
@@ -148,6 +152,20 @@ $defs:
...
+      - object
+    additionalProperties: false
+    properties:
+      root:
+        $ref: "#/$defs/Sampler"
+        description: |
+          Configure the wrapped sampler which provides the original sampling
+          decision that always_record modifies. DROP decisions are converted
+          to RECORD_ONLY, so spans are passed to processors without being
+          exported.
```
jack-berg:
```suggestion
          Configure the wrapped sampler which provides the original sampling
          decision that AlwaysRecord modifies. DROP decisions are converted
          to RECORD_ONLY, allowing processors to see all spans without sending them to exporters. 
```

snippets/Sampler_always_record_typical.yaml
```diff
@@ -0,0 +1,17 @@
+file_format: "1.1"
+
+tracer_provider:
+  processors:
+    - simple:
+        exporter:
+          console:
+  sampler:
+    # SNIPPET_START
+    # record all spans so processors see them (e.g. for span-to-metrics processing),
```
jack-berg: Hmm not sure this is a good snippet because the configured processor doesn't benefit from being able to see all spans. The span-to-metrics processing example is a good one, but there there is no built-in span-to-metric processor to demonstrate in the snippet. 

Could still be worth retaining if you can clarify what the always_record sampler allows while adding a caveat that leveraging this capability requires processors which are not demonstrated in the snippet.

jack-berg (commented):
Couple of small comments, but looks good!

schema/tracer_provider.yaml
dlanov: Applied the suggested wording and regenerated the compiled schema. Thank you.

snippets/Sampler_always_record_typical.yaml
dlanov: Clarified that the snippet does not configure a processor that benefits from observing all RECORD_ONLY spans, and that the wrapped sampler still controls which spans are exported.

jack-berg (approved):

marcalff (approved):
LGTM, thanks for the contribution.

PR #700, opened 2026-07-15
feat: add OTLP request and response size limits

### LINKED ISSUE #695: Add option for OTLP http/grpc request size limit
The OTLP proto spec now states that clients should apply request size limits, which are configurable and default to 64mb: https://github.com/open-telemetry/opentelemetry-proto/pull/782

We should extend [OtlpGrpcExporter](https://opentelemetry.io/docs/specs/otel-config/types/#type-otlpgrpcexporter), [OtlpHttpExporter](https://opentelemetry.io/docs/specs/otel-config/types/#type-otlphttpexporter) (and metric equivalents) to have new properties for configuring the size limit.

The [opentelemetry-specification/specification/protocol/exporter.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) document doesn't mention this yet and doesn't describe an env var for configuring it, so you might say that declarative config is blocked on adding new configuration properties until that exporter.md is updated.

However, we also say that the OTLP proto spec _is_ part of the specification - it just lives in a different repository. And so from that argument, there already is a specification saying that the request size should be configurable.

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-configuration/issues/695

Add configuration for limits introduced in:
- https://github.com/open-telemetry/opentelemetry-proto/pull/781
- https://github.com/open-telemetry/opentelemetry-proto/pull/782

From https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md

gRPC request:

> The client SHOULD limit the size of the request message, including before compression, to avoid overwhelming the server. It is RECOMMENDED to use 64 MiB as the default limit. Implementations SHOULD allow this limit to be configured.

gRPC response:

> The client MUST enforce a message size limit when receiving the response, including after decompression, to mitigate possible excessive memory usage caused by a misconfigured or malicious server. gRPC client implementations typically enforce a default incoming message size limit of 4 MiB, which is acceptable to use. Implementations SHOULD allow this limit to be configured.

HTTP request:

> The server MUST limit the size of the request body when parsing it, including after decompression, to mitigate possible excessive memory usage caused by a misconfigured or malicious client sending an oversized request. It is RECOMMENDED to use 64 MiB as the default limit. Implementations SHOULD allow this limit to be configured.

HTTP response:

> The client MUST limit the size of the response body when parsing it, including after decompression, to mitigate possible excessive memory usage caused by a misconfigured or malicious server. It is RECOMMENDED to use 4 MiB as the default limit. Implementations SHOULD allow this limit to be configured.


marcalff (commented):
LGTM.

The only question at this point is whether the new properties should be marked stable already, or named xxx/development.

Given issue #689, using plain names seem better.

schema/common.yaml
```diff
@@ -83,6 +83,24 @@ $defs:
...
           Known values include: gzip, none. Implementations may support other compression algorithms.
         defaultBehavior: none is used
+      max_request_size:
+        type:
+          - integer
+          - "null"
+        minimum: 0
+        description: |
+          Configure the maximum size of each export request body in bytes, before compression.
+          Value must be non-negative. A value of 0 indicates no limit (infinity) and is not recommended.
```
jack-berg, line 93: Similar to my comment [here](https://github.com/open-telemetry/opentelemetry-specification/pull/5186#discussion_r3668217485), I think it would be prudent to start off rejecting the 0 value (rather than interpreting it to mean infinity) and expand later to accept it if needed.

The current cases of "0 means infinity" are all timeout properties. This represents an expansion of the special value semantic outside of timeout.

schema/common.yaml
```diff
@@ -83,6 +83,24 @@ $defs:
           Configure compression.
           Known values include: gzip, none. Implementations may support other compression algorithms.
         defaultBehavior: none is used
+      max_request_size:
```
jack-berg, line 86: I'm not sure whether which spec we should hold ourselves to for which OTLP client properties should be configurable. 

Is it the [opentelemetry-proto/docs/specification.md](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md) or [opentelemetry-specification/specification/protocol/exporter.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) or some combination of both?

Its hard to read the [Configuration Options](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) of `opentelemetry-specification/specification/protocol/exporter.md` as any other than the source of truth for which configuration SDK OTLP exporters should have.

jack-berg: PR to extend `opentelemetry-specification/specification/protocol/exporter.md` with these new options: https://github.com/open-telemetry/opentelemetry-specification/pull/5235

jack-berg:
Seems like this PR was closed automatically as the result of https://github.com/open-telemetry/opentelemetry-specification/pull/5235 being merged. Not sure why 🤔.

Anyway, re-opening and I think this is unblocked.

schema/common.yaml
pellared: It allows `0` (infinity) as the request size limit is optional per spec

> The client SHOULD limit the size of the request 

See https://github.com/open-telemetry/opentelemetry-proto/pull/782

Users may want to as not applying the limit may improve the performance. Here is an example for OTel Go: https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp#WithMaxRequestSize

schema/common.yaml
pellared: I think this can be resolved :wink: Thanks!

jack-berg (approved):

schema/common.yaml
jack-berg: I don't see anything in the spec about interpreting zero as infinity. Are you saying that zero as infinity embodies the "SHOULD" keyword? 

I think its fine to go with zero as infinity. I don't feel too strongly either way. My point was just that this is an expansion of having special mechanics for the zero value. Up till this PR, the only places we've used zero as infinity are timeouts.

pellared: > Are you saying that zero as infinity embodies the "SHOULD" keyword?

Yes.

marcalff (approved):
LGTM.

reyang (approved):

PR #715, opened 2026-08-05
schema: add sql_commenter/development to ExperimentalDbInstrumentation

### LINKED ISSUE #619: Enable sql commenter via override
In [semantic conventions](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#sql-commenter) it mentions that:

> The instrumentation SHOULD allow users to pass a propagator to overwrite the global propagator. If no propagator is provided by the user, instrumentation SHOULD use the global propagator.

Reading this, we should have a `use_sqlcommenter` option within the db instrumentation options in declarative config with a default value of false.

This default state is defined via:

> SQL commenter-based context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.

om7057:
## Summary

Fixes #619.

This PR adds support for configuring SQL commenter-based trace context propagation in the configuration schema by introducing:

```yaml id="6pabv4"
instrumentation/development:
  general:
    db:
      sql_commenter/development: true
```

The property is defined under `general.db` since SQL commenter behavior is specified by the OpenTelemetry semantic conventions rather than being language-specific.

The `/development` suffix matches the current semantic convention status, which recommends that SQL commenter propagation remain an explicit opt-in rather than being enabled by default.

## Changes

* Added `sql_commenter/development` to `ExperimentalDbInstrumentation` in `schema/instrumentation.yaml`.
* Regenerated `opentelemetry_configuration.json`.
* Updated `snippets/ExperimentalInstrumentation_kitchen_sink.yaml`.
* Regenerated `language-support-status.md`.
* Added a CHANGELOG entry under **Unreleased**.

## Validation

* Ran `npm run compile-schema`.
* Verified the generated schema with `npx ajv-cli compile`.
* Ran `npm run validate-snippets`.
* Validated the existing example configurations against the regenerated schema.

This is a schema-only change and does not introduce any runtime or public API changes.


schema/instrumentation.yaml
```diff
@@ -278,6 +278,15 @@ $defs:
 
           See database migration: https://opentelemetry.io/docs/specs/semconv/database/
         defaultBehavior: uses the general stability_opt_in_list setting, or instrumentations continue emitting their default semantic convention version if not set
+      sql_commenter/development:
```
jack-berg: This probably better aligns with boolean property naming elsewhere in the repo.
```suggestion
      sql_commenter_enabled/development:
```

jack-berg (commented):
We'll want the approval of folks from @open-telemetry/semconv-db-approvers

schema/instrumentation.yaml
om7057: Updated to sql_commenter_enabled/development to match target_info_enabled/development and scope_info_enabled

om7057:
> We'll want the approval of folks from @open-telemetry/semconv-db-approvers

@jack-berg Thanks for flagging. Waiting on approval from team!

PR #724, opened 2026-08-18
feat: add experimental enabled to view stream

dashpole:
Implements View enabled configuration in support of open-telemetry/opentelemetry-specification#4809.
- Adds experimental ViewStream.enabled/development property
- Updates language support status matrices
- Adds snippets demonstrating enabled/development usage


language-support-status.md
```diff
@@ -206,7 +206,7 @@ Latest supported file format: `1.0.0`
 | `TracerProvider` | unknown |  | * `id_generator`: unknown<br>* `limits`: unknown<br>* `processors`: unknown<br>* `sampler`: unknown<br>* `tracer_configurator/development`: unknown<br> |
 | `View` | unknown |  | * `selector`: unknown<br>* `stream`: unknown<br> |
 | `ViewSelector` | unknown |  | * `instrument_name`: unknown<br>* `instrument_type`: unknown<br>* `meter_name`: unknown<br>* `meter_schema_url`: unknown<br>* `meter_version`: unknown<br>* `unit`: unknown<br> |
-| `ViewStream` | unknown |  | * `aggregation`: unknown<br>* `aggregation_cardinality_limit`: unknown<br>* `attribute_keys`: unknown<br>* `description`: unknown<br>* `name`: unknown<br> |
+| `ViewStream` | unknown |  | * `aggregation`: unknown<br>* `aggregation_cardinality_limit`: unknown<br>* `attribute_keys`: unknown<br>* `description`: unknown<br>* `name`: unknown<br>* `enabled/development`: not_implemented<br> |
```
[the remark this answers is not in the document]
dashpole: It is not_implemented because the feature hasn't been introduced in the spec yet

schema/meta_schema_language_go.yaml
```diff
@@ -237,7 +237,9 @@ typeSupportStatuses:
     propertyOverrides: []
   - type: ViewStream
     status: unknown
-    propertyOverrides: []
+    propertyOverrides:
+      - property: enabled/development
+        status: not_implemented
```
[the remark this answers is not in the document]
dashpole: I think it makes sense for this to be not_implemented, given we know no language has implemented it... I'll defer to reviewers.

snippets/View_enable_metric.yaml
```diff
@@ -0,0 +1,15 @@
+file_format: "1.1"
+
+meter_provider:
+  readers:
+    - periodic:
+        exporter:
+          console:
+  views:
```
jack-berg, line 9: Probably want to use the new `view_matching_mode/development: merge_all` now that #666 is merged.

jack-berg (approved):
Approve of the shape. Still pending corresponding spec approval / merge.

schema/meter_provider.yaml
```diff
@@ -630,6 +630,13 @@ $defs:
         description: |
           Configure attribute keys retained in the resulting stream(s).
         defaultBehavior: all attribute keys are retained
+      enabled/development:
```
dashpole: `enabled` is documented as an instrument-level switch but is a per-stream setting; multi-View and `viewmatchingmode` resolution is undefined. Scope the description to the stream and state the multi-View rule explicitly, e.g.

dashpole: Overlaps `aggregation: drop` with no stated precedence; the conflicting combination validates silently. As in F1 — one clause in the description ("`false` takes precedence over `.aggregation`; `true` does not override an explicit `drop` aggregation").

schema/meter_provider.yaml
```diff
@@ -630,6 +630,13 @@ $defs:
...
+        type:
+          - boolean
+          - "null"
+        description: |
+          Configure whether the instrument is enabled or not.
+        defaultBehavior: "it is determined by whether the instrument is opt-in: false if the instrument is opt-in, true otherwise"
```
dashpole: `defaultBehavior` hinges on "opt-in", a term that exists nowhere in the merged spec or in this schema, with no link. Reference the advisory parameter by its spec name and link it once #4809 merges (the `lychee`/`flint` lint job resolves links, so the anchor must exist): `defaultBehavior: the instrument's OptIn advisory parameter determines whether the stream is produced (see .../metrics/api.md#instrument-advisory-parameters)`.

schema/meter_provider.yaml
```diff
@@ -630,6 +630,13 @@ $defs:
...
   Aggregation:
```
dashpole: `defaultBehavior`'s "true otherwise" overclaims. State the delegation rather than the outcome — see F3's sketch.

schema/meter_provider.yaml
dashpole: Two different `enabled` axes now exist for metrics, with silent, undocumented precedence. One sentence on the new property: "A stream is only produced if the Meter that created the instrument is also enabled (see `.meter_configurator/development`).".

snippets/View_kitchen_sink.yaml
```diff
@@ -42,4 +42,5 @@ meter_provider:
           included:
             - foo.*
           excluded:
-            - foo.bar
\ No newline at end of file
+            - foo.bar
```
dashpole: Snippet coverage only demonstrates the easy direction, and the outstanding review comment cannot be actioned yet. Add `Viewdisablemetric.yaml` (or similar) showing `enabled/development: false`, and once #666 lands, a match-all snippet paired with the appropriate matching mode plus a comment explaining why the mode is needed there.

schema/meter_provider.yaml
```diff
@@ -630,6 +630,13 @@ $defs:
         description: |
           Configure attribute keys retained in the resulting stream(s).
         defaultBehavior: all attribute keys are retained
+      enabled/development:
+        type:
+          - boolean
+          - "null"
+        description: |
```
dashpole: Description wording drifts from the type's established style, and carries no spec link. Two small things, both already covered above: `Configure whether the instrument is enabled or not.` should be stream-scoped like its five siblings (F1), and unlike the neighbouring `aggregation`/`selector` descriptions it links nothing (F3).

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
```
dashpole: No negative validator fixture, and CONTRIBUTING's "update an example" rule is not literally met. #666 added `validator/shelltests/invalidviewmatchingmode.{yaml,test}` pinning the validator's real error text.

dashpole: Minor mechanical nits

snippets/View_enable_metric.yaml
dashpole: addressed in f159bf1

snippets/View_enable_metric.yaml
```diff
@@ -0,0 +1,16 @@
+file_format: "1.1"
+
+meter_provider:
+  view_matching_mode/development: composable
```
dashpole: `view_matching_mode/development` is above `# SNIPPET_START`, so per CONTRIBUTING.md:499 it isn't part of the demonstrated `View` snippet — and it can't be moved in, since it's a `MeterProvider` property (relocating it below the marker fails snippet validation with `must NOT have additional properties` against `#/$defs/View`). With a single registered View it's also a no-op, since `independent` and `composable` are identical for one matching View. Worth either dropping it, or showing this in a `MeterProvider_*` snippet with two Views that disagree on `enabled` — `MeterProvider_composable_views.yaml` from #666 already has the right layout, with `SNIPPET_START` above `view_matching_mode`.

PR #739, opened 2026-09-05
Read the declared `file_format` in the validator

### LINKED ISSUE #738: Validator ignores the `file_format` declared by a configuration file
## What happened

`otel_config_validator` never reads the `file_format` value a configuration file declares. Any value is accepted, including a version whose schema is nothing like the one the binary embeds.

Built from `main` (`3332c8b`) and run against a minimal configuration that differs only in `file_format`:

| `file_format` | result |
| --- | --- |
| `"1.1"` | accepted, exit 0 |
| `"0.1"` | accepted, exit 0 |
| `"99.7"` | accepted, exit 0 |
| `"banana"` | accepted, exit 0 |
| `1.1` (unquoted) | rejected: `expected string, but got number` |
| absent | rejected: `missing properties: 'file_format'` |

So exactly two things are checked about `file_format`: that it is present, and that it is a string. The value itself is never examined.

## What was expected

[VERSIONING.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md#file-format) states what an implementation should do when the version a file declares does not line up with the version the implementation targets:

> The `file_format` `MINOR` version is greater than the implementation `MINOR` version: The implementation should detect and emit a warning [...]

> The `file_format` `MAJOR` version does not align with the implementation `MAJOR` version. The implementation should produce an error [...]

Neither the warning nor the error happens.

## Why it happens

The schema is copied into the validator at build time by [`validator/Makefile`](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/validator/Makefile#L35-L39), embedded via `//go:embed schema/*`, and compiled from a hardcoded URL in [`validator/main.go`](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/validator/main.go#L161):

```go
schema, err := c.Compile("https://opentelemetry.io/otelconfig/opentelemetry_configuration.json")
```

The binary therefore has no record of which version of the format its embedded schema describes, and nothing compares that to what the file declares.

## Side effect: the shelltest fixtures have drifted

Because the value is never read, stale versions in the fixtures went unnoticed. On `main`, `validator/shelltests/hex_integer.yaml`, `string_for_int.yaml` and `schema_option.yaml` all declare `file_format: "0.1"`, and `json_out.test` and `yaml_out.test` run with `FILE_FORMAT=0.1` and assert `"0.1"` in the validator's output.

The release step that keeps this in step, `make update-file-format` (see [RELEASING.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/RELEASING.md)), rewrites `file_format` in `./examples` and `./snippets` only, not in `./validator/shelltests`.

## Suggested fix

Record the file format version the embedded schema describes, compare it to the declared value, and follow the rules already written in `VERSIONING.md`: error on a differing `MAJOR`, warn on a newer `MINOR`, error on a value that is not `MAJOR.MINOR` with an optional meta tag.

ocelotl:
Fixes #738.

`otel_config_validator` never read the `file_format` a configuration file declares, so a file claiming `0.1`, `99.7` or `banana` validated against the current schema and exited 0. This makes it read that value and follow the rules already written in [VERSIONING.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/VERSIONING.md#file-format).

## Behaviour

| `file_format` | before | after |
| --- | --- | --- |
| `"1.1"` | accepted | accepted |
| `"1.9"` | accepted silently | accepted, warns that it is newer than the embedded schema |
| `"1.1-rc.2"` | accepted | accepted, the meta tag is discarded |
| `"0.1"` | accepted silently | rejected, exit 1 |
| `"99.7"` | accepted silently | rejected, exit 1 |
| `"banana"` | accepted silently | rejected, exit 1 |

A newer `MINOR` is a warning on stderr and still exits 0, so a file written for a later release keeps validating.

## Commits

* `Record which file format version the embedded schema describes` adds the `supportedFileFormat` constant, plus a test that fails if it drifts from what `./examples` declare.
* `Reject a configuration whose major version the schema does not describe` adds the check and wires it in after schema validation, so existing schema errors still surface first.
* `Bring the shelltest fixtures onto a version the schema describes` updates fixtures that had drifted to `0.1` and adds three shelltests for the new outcomes.
* `Document the version check and the constant a release has to bump` covers `validator/README.md` and `RELEASING.md`.

## A note on the fixtures

Four fixtures declared `file_format: "0.1"` and two asserted `"0.1"` in the validator's output. They passed only because the value was never read. `env_prefix.test` also needed `FILE_FORMAT=1.1`, because with the variable unset the field expanded to an empty string, which is a string and so satisfied the schema.

These drifted because `make update-file-format` rewrites `./examples` and `./snippets` but not `./validator/shelltests`. Rather than extend that target, this adds `TestSupportedFileFormatMatchesExamples`, so `go test` in `./validator` fails whenever the constant and the examples disagree, and `RELEASING.md` now says the constant has to be bumped alongside the rest.

## Verification

* `go vet ./...` clean.
* `go test .` passes, including `TestSplitFileFormat`, `TestCheckFileFormat` and `TestSupportedFileFormatMatchesExamples`.
* `make validator-run-shelltests` passes 19 of 19, 0 failures.

## Out of scope

This does not let the validator hold more than one schema version. It records the single version it embeds and reports honestly when a file does not match it. Serving several versions from one binary would be a larger change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 * Add `max_request_size` and `max_response_size` to OTLP exporters
   ([#700](https://github.com/open-telemetry/opentelemetry-configuration/pull/700))
 
+### Tooling
+
+* Validator: read the `file_format` a configuration file declares and check it
+  against the schema version the binary embeds
+  ([#739](https://github.com/open-telemetry/opentelemetry-configuration/pull/739))
+
 ## v1.1.0 - 2026-06-05
 
 ### Schema
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -43,3 +43,7 @@ make update-file-format
 ```
 
 Merge a PR with the changes to `main`.
+
+Also update `supportedFileFormat` in [validator/main.go](./validator/main.go) to the same
+version, so the validator keeps rejecting configuration it cannot describe. `go test` in
+`./validator` fails while the two disagree.
--- a/validator/README.md
+++ b/validator/README.md
@@ -64,6 +64,23 @@ Environment variable substitution is supported with the syntax `${VARIABLE}` or
 found in the
 [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/data-model.md#environment-variable-substitution).
 
+#### File Format Version
+
+The embedded schema describes one version of the file format, and the validator
+compares it against the `file_format` a configuration file declares, following
+[VERSIONING.md](../VERSIONING.md#file-format):
+
+- A different `MAJOR` version is an error, because the embedded schema does not
+  describe that version of the format.
+- A newer `MINOR` version is a warning, because the configuration may use
+  properties the embedded schema does not contain.
+- A `file_format` that is not `MAJOR.MINOR`, optionally followed by a meta tag
+  such as `1.0-rc.2`, is an error.
+
+Passing `-s`/`--schema` does not change the version the validator expects, so a
+schema for a different `MAJOR` version has to be used with a build of the
+validator that targets it.
+
 #### Using a Different Schema Version
 
 To use a version of of the schema other than the one builtin to the
--- a/validator/main.go
+++ b/validator/main.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"embed"
 	"encoding/json"
+	"fmt"
 	jsonschema "github.com/santhosh-tekuri/jsonschema/v5"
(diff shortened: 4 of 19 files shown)


marcalff (approved):
LGTM, thanks for the fix.

codeboten (approved):

PR #751, opened 2026-09-14
chore(validator): open a PR to bump validator version on schema release

tsloughter:
I'd like to see the validator get a version bump and embed the latest schema on each schema version release. This PR is an attempt to do so but it isn't completely streamlined.

On a schema release it will do the necessary version bumps in the validator and open a PR to merge those in. But then someone has to manually push a tag for these changes after the PR has been merged.

Now that I think about it maybe actions can trigger after the PR is merged to do the tag... I'll look into that in a minute and add to this PR if it is possible.

Note: the code was Codex generated but I reviewed it.


jack-berg:
Are you using the validator in erlang? Do any @open-telemetry/configuration-approvers know of additional validator use elsewhere in the otel project?

Trying to figure out who the stakeholders are for this type of change

